PR #9, opened 2017-05-14
Adds better commentary to lucky-strike. Adds hit-harder.

cormacmccarthy:


tweettypography:
Can we move the stroke of luck / curse logic to a shared place? It varies somewhat per card but at least to the base class of the card type? It's duplicated quite a bit right now

cormacmccarthy:
Good call

cormacmccarthy:
Done

Issue #16, opened 2017-05-15
Card Energy


cormacmccarthy:
Each turn, your monster should gain N amount of energy (dependent upon level), of which unused energy will carry over from turn to turn, round to round

Each card has a unique energy requirement. Monsters can play as many cards as they have energy for each turn.

Rounds consist of 3 turns.

Decks consist of 7 cards. Even if you have energy left at the end of the round, if you have played 7 cards, your energy just carries over to the next round.

Once we have #15 Card Chaining, consider differentiating between caps on Action vs Logic cards. Say up to 7 Action Cards and up to 7 Logic Cards, with a max of 10 Cards, so you can do something like this:

5E Hit
1E Skip next if Bloodied
5E Hit
1E Skip next if Healthy
4E Heal
1E Skip next if Healthy
1E Skip next if Opponent Bloodied
3E Flee
4E Heal
8E Pound

For a monster with 7E per turn (21E per round), as long as they were not bloodied, their round's  could play out like this:

Round 1:
Turn 1, 7E:
-5E Hit
-1E Skip
Turn 2, 8E:
-5E Hit
-1E Skip
Turn 3, 9E:
-1E Skip
-1E Skip
-4E Heal

Round 2:
Turn 1, 10E:
-5E Hit
-1E Skip
Turn 2, 11E:
-5E Hit
-1E Skip
-1E Skip
-1E Skip
Turn 3, 9E:
-4E Heal

Round 3:
Turn 1, 12E:
-5E Hit
-1E Skip
-5E Hit
-1E Skip
Turn 2, 7E:
-1E Skip
-1E Skip
-4E Heal
Turn 3, 8E:
**-8E Pound**

This lets you do things like squeeze in an extra bonus card that you get to play every 3 rounds or so that normally takes up too much energy for you to want to risk it in your regular rotation, but can change the game when your energy builds up for it.

cormacmccarthy:
The game has evolved a lot since this issue was opened. I think there's some good stuff here, but I'm moving away from the idea of card chaining.

I still like energy because it lets you play some smaller cards without them taking up a ton of space (curse, bless, flee) and leaves the concept of epic cards open.

[...]

PR #28, opened 2017-05-18
makes decks always be 7 cards

cormacmccarthy:
relates to #16

### DIFF
--- a/monsters/base.js
+++ b/monsters/base.js
@@ -25,12 +25,11 @@ class BaseMonster extends BaseCreature {
 		});
 	}
 
-	get cardSlots () {
-		if (this.options.cardSlots === undefined) this.cardSlots = DEFAULT_CARD_SLOTS;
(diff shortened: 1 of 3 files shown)


monsters/base.js
```diff
@@ -27,10 +27,10 @@ class BaseMonster extends BaseCreature {
 
 	get cardSlots () {
 		if (this.options.cardSlots === undefined) this.cardSlots = DEFAULT_CARD_SLOTS;
+		if (this.options.cardSlots !== DEFAULT_CARD_SLOTS) this.cardSlots = DEFAULT_CARD_SLOTS; // enforce default for now
 		const cardSlots = this.options.cardSlots || 0;
-		const level = this.level || 0;
 
-		return cardSlots + level;
+		return cardSlots;
```
tweettypography: If we're enforcing the default for now why not just have `get cardSlots ()` return `DEFAULT_CARD_SLOTS` instead managing an option at all?

cormacmccarthy:
Done

PR #30, opened 2017-05-18
removes attack dice modifier. Adding a 1d4 to hitCheck is way too ove…

### LINKED ISSUE #29: hitCheck should not include +1d4 per level
This comes from the attackModifier. We already have an attackBonus of +1 per level. This coupled with the attackModifier practically makes AC irrelevant, as well as makes the different creature penalties and bonuses meaningless.

cormacmccarthy:
…r-powered

fixes #29

### DIFF
--- a/creatures/base.js
+++ b/creatures/base.js
@@ -159,17 +159,20 @@ Battles won: ${this.battles.wins}`;
 		return ac;
 	}
 
-	get bonusAttackDice () {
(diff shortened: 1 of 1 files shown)


tweettypography:
I would leave this in the cards and simply make the change you did to creatures. That way if we ever have a place where a bonus die roll is needed the hooks are already there.

cormacmccarthy:
Done.

PR #36, opened 2017-05-19
Gh31   end of battle xp and card should be shown to user

### LINKED ISSUE #31: End of battle XP and card should be shown to user
At the end of the battle, you gain XP and possibly a card. That should be more obvious.

cormacmccarthy:
fixes #31 

image

I'm going to bet that this is going to be more of a starting point. I don't quite grok the new channelManager yet, and I'm betting I'm sorely abusing or underutilizing either it or the emit system or both...

### DIFF
--- a/game.js
+++ b/game.js
@@ -96,8 +96,7 @@ class Game extends BaseClass {
 		this.channelManager.queueMessage({
 			announce: `${creature.identity} gained ${xpGained}XP`,
 			channel,
-			channelName,
-			event: { name: 'gainedXP', properties: { creature } }
+			channelName
 		});
 	}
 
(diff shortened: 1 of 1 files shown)


game.js
```diff
@@ -81,11 +81,45 @@ class Game extends BaseClass {
...
 
+	announceXPGain (className, game, { contestant, creature, xpGained }) {
+		const channel = contestant.channel;
+		const channelName = contestant.channelName;
+
+		this.channelManager.queueMessage({
+			announce: `${creature.identity} gained ${xpGained}XP`,
+			channel,
+			channelName,
+			event: { name: 'gainedXP', properties: { creature } }
```
tweettypography: You don't have to emit an event here unless you want it for something else - this event prop is basically to allow for events that only fire off _after_ the message has actually been displayed to the user. So it's basically an event to notify the system that a message has been displayed.

cormacmccarthy: Oh! I wondered what this event prop was for! Never got around to following it through the pipeline to see what happened to it... Thanks!

cormacmccarthy:
I can't remember what my questions about it were... If I seem to be using it right and putting things in a reasonable place, then I guess I'm good for now. If I have a specific question later I'll ask :)

PR #37, opened 2017-05-19
Gh24   adds call monster from ring

### LINKED ISSUE #19: Allow removal from Ring
need a way to remove yourself from the ring (the regular ring) as long as you're not already in an encounter. And maybe even a delay to give you time to get out after someone joins if you see they're way more powerful than you...

That delay could also allow us to let third or fourth players jump in and join a battle before it starts, might be fun

cormacmccarthy:
Fixes #19

### DIFF
--- a/characters/beastmaster.js
+++ b/characters/beastmaster.js
@@ -144,15 +144,20 @@ Which monster would you like to ${action}?`,
 		let monster = monsterName;
 		// If monster is an empty object or string, it means they didn't pass anything through
 		if (!monster || Object.keys(monster).length === 0) {
-			monster = ring.contestants.find(contestant => contestant.character.name === character.name).monster;
+			console.log('here');
+			monster = ring.contestants.find(contestant => contestant.character.name.toLowerCase() === character.name.toLowerCase()).monster;
+			console.log('monster gotten', monster);
 		}
-		const monsterInRing = ring.contestants.filter(contestant => contestant.monster.name === monster.name);
+		const monsterInRing = ring.contestants.find((contestant) => {
+			console.log(contestant.monster.givenName.toLowerCase(), monster.toLowerCase());
+			return contestant.monster.givenName.toLowerCase() === monster.toLowerCase();
+		}).monster;
 
 		return Promise
 			.resolve()
 			.then(() => {
-				if (monsterInRing && monsterInRing.length > 0) {
-					return ring.removeMonster(monster, character, channel, channelName);
+				if (monsterInRing && monsterInRing.givenName) {
+					return ring.removeMonster(monsterInRing, character, channel, channelName);
 				}
 
 				return Promise.reject(channel({
--- a/characters/beastmaster.js
+++ b/characters/beastmaster.js
@@ -152,7 +152,7 @@ Which monster would you like to ${action}?`,
 		return Promise
 			.resolve()
 			.then(() => this.chooseMonster({ channel, monsters, monsterName, action: 'call from the ring', reason: 'they do not appear to be in the ring.' })) // eslint-disable-line max-len
-			.then(monsterInRing => ring.removeMonster({ monsterInRing, character: this, channel, channelName }));
+			.then(monsterInRing => ring.removeMonster({ monster: monsterInRing, character: this, channel, channelName }));
 	}
 
 	sendMonsterToTheRing ({ monsterName, ring, channel, channelName }) {
--- a/characters/beastmaster.js
+++ b/characters/beastmaster.js
@@ -141,15 +141,7 @@ Which monster would you like to ${action}?`,
 	}
 
 	callMonsterOutOfTheRing ({ monsterName, ring, channel, channelName }) {
-		const character = this;
-		const monsters = [];
-
-		// filter down to monsters already in the ring
-		ring.contestants.forEach((contestant) => {
-			if (contestant.character.name === character.name) {
-				monsters.push(contestant.monster);
-			}
-		});
+		const monsters = ring.getMonsters(this);
 
 		if (monsters.length <= 0) {
 			return Promise.reject(channel({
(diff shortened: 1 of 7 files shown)


characters/beastmaster.js
```diff
@@ -139,6 +140,29 @@ Which monster would you like to ${action}?`,
...
+	callMonsterOutOfTheRing ({ monsterName, ring, channel, channelName }) {
+		const character = this;
+		const monsters = [];
+
+		// filter down to monsters already in the ring
+		ring.contestants.forEach((contestant) => {
+			if (contestant.character.name === character.name) {
+				monsters.push(contestant.monster);
+			}
+		});
```
tweettypography: You should probably use filter for this, not foreach. Also - we don't have a method that gets the monsters for a particular contestant's name out of the ring yet? We probably should.

And one more thing -- you don't actually even have to compare names because contestants would be a direct reference to each other.

ring/index.js
```diff
@@ -31,6 +31,15 @@ class Ring extends BaseClass {
 		return this.options.contestants || [];
 	}
 
+	getMonsters (targetCharacter) {
+		let targetContestants = [...this.contestants];
```
tweettypography: You can skip this step of duplicating the array, neither `filter` nor `map` is destructive.

ring/index.js
```diff
@@ -31,6 +31,15 @@ class Ring extends BaseClass {
 		return this.options.contestants || [];
 	}
 
+	getMonsters (targetCharacter) {
```
tweettypography: Ah, it's the `getAllMonsters` method (of `Game`) that I was thinking of. So this one is the equivalent `getMonsters` that gets all monsters in the ring. Makes sense. Do you think there's any reason why we might want to use the object-style return that's in the other method? I'm fine with either way.

tweettypography: * that is, all monsters in the ring, optionally filtered by character

ring/index.js
```diff
@@ -116,8 +125,7 @@ class Ring extends BaseClass {
 	}
 
 	monsterIsInRing (monster) {
```
tweettypography: Should we change this to optionally take the character as well and use your new method from above -- allowing you to check whether the monster is in the ring, or specifically sent by a specific character into the ring?

ring/index.js
```diff
@@ -31,6 +31,15 @@ class Ring extends BaseClass {
...
+	getMonsters (targetCharacter) {
+		let targetContestants = [...this.contestants];
+		if (targetCharacter) {
+			targetContestants = targetContestants.filter(contestant => contestant.character === targetCharacter);
+		}
+
+		return targetContestants.map(contestant => contestant.monster);
+	}
+
 	removeMonster (monster, character, channel, channelName) {
```
tweettypography: We mostly use the ({ object, destructuring, syntax }) for our multi-argument methods. Is there a reason why this one is different?

characters/beastmaster.js
```diff
@@ -160,7 +152,7 @@ Which monster would you like to ${action}?`,
 		return Promise
 			.resolve()
 			.then(() => this.chooseMonster({ channel, monsters, monsterName, action: 'call from the ring', reason: 'they do not appear to be in the ring.' })) // eslint-disable-line max-len
-			.then(monsterInRing => ring.removeMonster(monsterInRing, character, channel, channelName));
```
tweettypography: This was actually somewhat more clear as a named variable...

ring/index.js
cormacmccarthy: There isn't really a use-case for that right now. Someday if we want to make a "whose monster is this?" method, or an `isOwnedBy` method or something, that'd be fine, but there's no need for it yet.

ring/index.js
cormacmccarthy: It's probably better if they both return arrays since thats how the monsters are currently stored. I'll change the other one for consistency.

ring/index.js
cormacmccarthy: Just habit :)

cormacmccarthy:
I have a bug. Didn't mean to push what I had yet, don't merge...

ring/index.js
tweettypography: The lookup by name feature of the other one is actually used in the code, which is why it is structured the way it is. If you're going to change it you'll have to make sure to update any places that use it.

cormacmccarthy:
OK, @tweettypography all ready for you.

monsters/index.js
```diff
@@ -25,7 +25,8 @@ const spawn = (channel, { type, name, color, gender, cards, game } = {}) => {
 
 	let monsterNames = [];
 	if (game) {
-		monsterNames = game.getAllMonsters().reduce((monsters, monster) => monsters.push(monster.givenName), []);
+		const allMonsters = game.getAllMonsters();
+		monsterNames = allMonsters.reduce((monsters, monster) => monsters.concat(monster.givenName), []);
```
tweettypography: Looks like we're using a `reduce` for a `map`! Isn't this just `allMonsters.map(monster => monster.givenName)`?

game.js
```diff
@@ -449,33 +465,27 @@ ${monsterCard(monster)}`
...
 
 	getAllMonsters () {
-		return reduce(this.characters, (obj, character) => {
-			character.monsters.forEach((monster) => {
-				obj[monster.givenName.toLowerCase()] = monster;
-			});
-
-			return obj;
-		}, {});
+		return reduce(this.characters, (all, character) => all.concat(character.monsters), []);
```
tweettypography: I feel like all the code using this was cleaner before - we're having to do a bunch of find loops with nested `toLowerCase` where we used to just do a lookup. Not a huge problem as long as it works, but are we sure this is better?

cormacmccarthy: Nope. I don't like the new code as much honestly. I more or less did this for consistency sake. This one was that only one that returned monsters as an Object instead of as an array (as it is normally stored).

tweettypography: What if we just renamed it to `getAllMonstersLookup`? We could always make a `getAllMonsters` still if we need it but that way it's clearer what you're getting.

monsters/index.js
```diff
@@ -25,7 +25,8 @@ const spawn = (channel, { type, name, color, gender, cards, game } = {}) => {
 
 	let monsterNames = [];
 	if (game) {
-		monsterNames = Object.keys(game.getAllMonsters());
+		const allMonsters = game.getAllMonsters();
+		monsterNames = allMonsters.reduce((monsters, monster) => monsters.concat(monster.givenName), []);
```
cormacmccarthy: But you like reduce _so much_! 

Seriously though, I'm still wrapping my head around reduce/map/etc. If left to my own devices, I always just use `forEach` (which I recognize is bad). #habitsabitch

game.js
cormacmccarthy: 🤔 I like it!

PR #38, opened 2017-05-21
Gh23   card pool display

cormacmccarthy:
fixes #23 
fixes #22 

Allows player to look at all of their cards, monsters, and monsters in ring.

Displays only a single instance of each card type in the equip menu, with counts of how many of each type the player has in brackets.

PR #39, opened 2017-05-22
Gh21   equip shortcut

### LINKED ISSUE #21: Equip shortcut
Need shorthand for quickly equipping a monster with cards

`equip legolost with hit, hit, heal, flee`

cormacmccarthy:
fixes #21 

Allows player to specify cards to equip monster with using comma separated list

depends on #38

### DIFF
--- a/game.js
+++ b/game.js
@@ -425,7 +425,16 @@ ${monsterCard(monster)}`
 						.catch(err => log(err));
 				},
 				equipMonster ({ monsterName, cardSelection } = {}) {
-					return character.equipMonster({ monsterName, cardSelection, channel })
+					let selectedCards = [];
+					if (cardSelection) {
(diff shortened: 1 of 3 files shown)


game.js
```diff
@@ -361,8 +357,8 @@ ${monsterCard(monster)}`
 					return character.spawnMonster(channel, Object.assign({}, options, { game }))
 						.catch(err => log(err));
 				},
-				equipMonster ({ monsterName } = {}) {
-					return character.equipMonster({ monsterName, channel })
+				equipMonster ({ monsterName, cardSelection } = {}) {
```
tweettypography: This should probably be an array, not a string - split it before it's passed instead

cormacmccarthy: Done

PR #47, opened 2017-06-14
Gh45   monsters should leave ring after battle

### LINKED ISSUE #45: Monsters should leave ring after battle
After a battle if you try and send a new monster into the ring, it says the ring is full.

cormacmccarthy:
fixes #45 

@dbb I don't know why the save function error is occurring, but I'm swallowing it for the meantime (because otherwise the game is completely broken after the first fight)

### DIFF
--- a/characters/beastmaster.js
+++ b/characters/beastmaster.js
@@ -2,6 +2,7 @@ const BaseCharacter = require('./base');
 const { spawn, equip } = require('../monsters');
 const { getMonsterChoices } = require('../helpers/choices');
 const TENSE = require('../helpers/tense');
+const { capitalize } = require('../helpers/capitalize');
 
 const { monsterCard } = require('../helpers/card');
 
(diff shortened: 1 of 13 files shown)


characters/beastmaster.js
```diff
@@ -231,10 +232,14 @@ Which monster would you like to ${action}?`,
...
 			})
-			.then(monster => channel({
-				announce: `${monster.givenName} has begun to revive.`
-			})
-				.then(() => monster));
+			.then((monster) => {
+				const timeToRevive = (monster.respawn() - Date.now()) / (60 * 60 * 1000);
+
+				return channel({
+					announce: `${monster.givenName} has begun to revive. ${capitalize(monster.pronouns[0])} is a level ${monster.level} monster, and therefore will be revived in ${timeToRevive} hours.`// eslint-disable-line max-len
```
tweettypography: This doesn't really handle the case where it's 0 or 1 hours.

characters/beastmaster.js
```diff
@@ -231,10 +232,14 @@ Which monster would you like to ${action}?`,
 
 				return monster;
 			})
-			.then(monster => channel({
-				announce: `${monster.givenName} has begun to revive.`
-			})
-				.then(() => monster));
+			.then((monster) => {
+				const timeToRevive = (monster.respawn() - Date.now()) / (60 * 60 * 1000);
```
tweettypography: There should probably be a helper function for this that can be shared between the actual logic and this display of the info.

game.js
```diff
@@ -56,7 +56,18 @@ class Game extends BaseClass {
...
+					const gameJSON = JSON.stringify(game);
+					// console.log('not circular', gameJSON);
+					return stateSaveFunc(gameJSON);
+				} catch (e) {
+					// console.log('err circular');
+					// console.log(e);
+					// console.log('game', game);
+				}
+				return false;
+			};
```
tweettypography: Wait... what's happening here? We should fix whatever is causing this issue rather than swallowing errors.

tweettypography:
@cormacmccarthy Updated to fix the bug

PR #44, opened 2017-06-11
Adds commands to players handbook

cormacmccarthy:
Adds command list to players handbook

Makes look at command more forgiving

PR #50, opened 2017-06-28
matches on dm guide and dmg for dungeon masters guide

cormacmccarthy:
addresses comment on #44

### DIFF
--- a/game.js
+++ b/game.js
@@ -554,7 +554,7 @@ ${monsterCard(monster)}`
 			}
 
 			// Is it a dungeon master guide?
-			if (thing.match(/dungeon master(?:s)? guide/i)) { // dmg will talk about how to make new cards, monsters, and dungeons. Basically, the developer docs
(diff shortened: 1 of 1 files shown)


game.js
```diff
@@ -554,7 +554,7 @@ ${monsterCard(monster)}`
 			}
 
 			// Is it a dungeon master guide?
-			if (thing.match(/dungeon master(?:s)? guide/i)) { // dmg will talk about how to make new cards, monsters, and dungeons. Basically, the developer docs
+			if (thing.match(/(?:dungeon master(?:s)?|dm) guide|^dmg$/i)) { // dmg will talk about how to make new cards, monsters, and dungeons. Basically, the developer docs
```
tweettypography: Why do you use explicitly match start and end for dmg but not for dungeon master guide?

cormacmccarthy: Just in case something like this: `look at monster badmgoo`

People _probably_ won't name their monster "dm guide" or "dungeon masters guide" though...

cormacmccarthy:
related to: https://github.exacttarget.com/developer-productivity/sf-mc--hubot-jane/pull/115

image

PR #53, opened 2017-11-16
Adds effects to the fight loop

tweettypography:
Effects allow arbitrary modifications during the battle. For example, a card could trigger the creation of an effect which watches for another card and then modifies the behavior of that card. Effects are applied in three phases:

1. Attack Phase - where each of the effects from the monster who is playing the card are applied
2. Defense Phase - where each of the effects from the monster who the card may be targeting are applied
3. Global Phase - where effects that are from the ring itself are applied

These three phases are followed by the actual play of the card.

And effect is a function with the following signature: `card:Card, player:Creature, target:Creature, ring:Ring, phase:constString`


creatures/base.js
```diff
@@ -189,13 +215,15 @@ Battles won: ${this.battles.wins}`;
...
 		}
+		damageModifier += this.conditions.damageModifier || 0;
 
 		return damageModifier;
 	}
 
 	get maxHp () {
 		let maxHp = this.options.maxHp || this.DEFAULT_HP;
 		maxHp += Math.min(this.level * 2, MAX_HP_BOOST); // Gain 2 hp per level up to the max
+		maxHp += this.conditions.maxHp || 0;
```
cormacmccarthy: this.conditions.maxHpMod or something like that may be more appropriate here. Just this.conditions.maxHp could lead one to believe that it was the actual maxHp rather than a modification to maxHp.

ring/index.js
```diff
@@ -266,11 +284,32 @@ class Ring extends BaseClass {
...
+				const nextMonster = contestants[nextContestant].monster;
+
+				// Now we're going to run through all of the possible effects
+				// Each effect should either return a card (which will replace the card that was going to be played)
+				// or do something in the background and then return nothing (in which case we'll keep the card we had)
+
+				// First, run through the effects from the current monster
+				card = monster.encounterEffects.reduce((result, effect) => effect({
+					card: result, player: monster, target: nextMonster, ring, phase: ATTACK_PHASE
+				}) || card, card);
```
cormacmccarthy: The way this is structured is not intuitive. Here's an uglier, less aesthetically satisfying structure that's much more readable:

```
				card = monster.encounterEffects.reduce((result, effect) => {
					const modifiedCard = effect({ card: result, player: monster, target: nextMonster, ring, phase: ATTACK_PHASE });

					return modifiedCard || card;
				}, card);
```

creatures/base.js
tweettypography: I based this off of the logic we already had in place for `ac`. So we should probably either rename all of them to `acModifier` `maxHpModifier` etc or leave them all like this. Another option would be to change `conditions` to `modifiers`. We could always add another object called `conditions` in the future if we wanted some conditions that aren't modifiers - like `frozen` `invisible` `poisoned` etc

Thoughts?

ring/index.js
tweettypography: Updated

creatures/base.js
tweettypography: I'm going with renaming `conditions` to `modifiers`

PR #54, opened 2017-11-17
 Updates the flee card to only flee if hp < 10

tweettypography:


cards/flee.js
```diff
@@ -10,39 +10,46 @@ class FleeCard extends BaseCard {
...
-						return resolve(player.leaveCombat(target));
-					}
-
-					this.emit('stay', {
-						fleeResult: fleeRoll.result,
-						fleeRoll,
-						player,
-						target
-					});
+			if (player.hp < 10) {
```
cormacmccarthy: Can we change this to be whether the player is bloodied? (half hp)

PR #71, opened 2017-12-07
Gh42   dead for negative bloodied

### LINKED ISSUE #42: "dead" for -bloodied
If you get a huge hit and knock the other monster to below negative 1/2 maxHP, they should be dead irretrievably.

This should be rare enough to almost never happen.

For level 1 monsters, the most they can hit for it about 14 on a natural 20. Most level 1 monsters maxHP is at least 24. In order for this to happen to a level 1 monster, they would have to have only 1HP left, and have their opponent roll at natural 20 to hit, and have their opponent have bonuses to hit.

cormacmccarthy:
fixes #42

### DIFF
--- a/creatures/base.js
+++ b/creatures/base.js
@@ -128,6 +128,24 @@ Battles won: ${this.battles.wins}`;
 		}
 	}
 
+	get bloodiedValue () {
+		return Math.floor(this.maxHp / 2);
+	}
+
+	get bloodied () {
+		return this.hp < this.bloodiedValue();
+	}
+
+	set bloodied (hp) {
+		this.setOptions({
+			hp: this.bloodiedValue()
+		});
+	}
+
+	get destroyed () {
+		return this.hp < -this.bloodiedValue()
+	}
+
 	get hp () {
 		if (this.options.hp === undefined) this.hp = this.maxHp;
 
@@ -283,10 +301,6 @@ Battles won: ${this.battles.wins}`;
 		const hp = this.hp - damage;
 		const originalHP = this.hp;
 
-		if (hp <= 0) {
-			this.hp = 0;
-		}
-
 		this.hp = hp;
 
 		this.emit('hit', {
@@ -349,9 +363,15 @@ Battles won: ${this.battles.wins}`;
 	}
 
 	die (assailant) {
-		this.emit('die', {
-			assailant
-		});
+		if () {
+			this.emit('destroy', {
+				assailant
+			});
+		} else {
+			this.emit('die', {
+				assailant
+			});
+		}
 
 		if (this.hp > 0) {
 			this.hp = 0;
--- a/creatures/base.js
+++ b/creatures/base.js
@@ -136,17 +136,17 @@ Battles won: ${this.battles.wins}`;
 	}
 
 	get bloodied () {
-		return this.hp < this.bloodiedValue();
+		return this.hp < this.bloodiedValue;
 	}
 
 	set bloodied (hp) {
 		this.setOptions({
-			hp: this.bloodiedValue()
+			hp: this.bloodiedValue
 		});
 	}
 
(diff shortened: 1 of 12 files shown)


creatures/base.js
```diff
@@ -286,10 +304,6 @@ Battles won: ${this.battles.wins}`;
 		const hp = this.hp - damage;
 		const originalHP = this.hp;
 
-		if (hp <= 0) {
-			this.hp = 0;
-		}
-
```
tweettypography: Why this logic even in here??

creatures/base.js
```diff
@@ -352,9 +366,17 @@ Battles won: ${this.battles.wins}`;
 	}
 
 	die (assailant) {
-		this.emit('die', {
-			assailant
-		});
+		if (this.hp <= 0) {
+			if (this.destroyed()){
```
tweettypography: Should be `this.destroyed` (not a function)

creatures/base.js
```diff
@@ -352,9 +366,17 @@ Battles won: ${this.battles.wins}`;
...
+				this.emit('destroy', {
+					assailant
+				});
```
tweettypography: You might want to just issue the same `die` even and include the parameter `destroyed` - would probably make it simpler to handle all of the cases where the die event might be listened to...

ring/index.js
```diff
@@ -29,8 +29,7 @@ class Ring extends BaseClass {
 			ring.battles.push(results);
 		});
 
-		const ring = this;
-		this.channelManager.on('win', (className, channel, { contestant }) => ring.handleWinner({ contestant }));
+		this.channelManager.on('win', (className, channel, { contestant }) => this.handleWinner({ contestant }));
 		this.channelManager.on('loss', (className, channel, { contestant }) => this.handleLoser({ contestant }));
 		this.channelManager.on('draw', (className, channel, { contestant }) => this.handleTied({ contestant }));
```
tweettypography: I don't see you actually listening for the `permaDeath` event here?

creatures/base.js
cormacmccarthy: oh weird... That's going to take some getting used to.

game.js
```diff
@@ -185,14 +186,21 @@ ${monsterCard(monster, contestant.lastMonsterPlayed !== monster)}`
...
 		const channel = this.publicChannel;
+		let announce;
 
-		channel({
-			announce:
-`${monster.identityWithHp} is killed by ${assailant.identityWithHp}
+		if (destroyed) {
+			announce = `${monster.identityWithHp} has been sent to the land of ${monster.pronouns} fathers by ${assailant.identityWithHp}
+
+			☠️  R.I.P ${monster.identity}
```
tweettypography: Did you mean for there to be this extra whitespace on the left?

ring/index.js
```diff
@@ -453,12 +453,21 @@ class Ring extends BaseClass {
 
 				if (contestant.monster.dead) {
 					contestant.lost = true;
-					this.channelManager.queueMessage({
-						announce: `${contestant.monster.givenName} has died in battle. You may now \`revive\` or \`dismiss\` ${contestant.monster.pronouns[1]}.`,
-						channel,
-						channelName,
-						event: { name: 'loss', properties: { contestant } }
-					});
+					if (contestant.monster.destroyed()) {
```
tweettypography: I still think this should be `if (contestant.monster.destroyed) {` - I'm kind of confused how it's working...

game.js
cormacmccarthy: Not originally, but I kinda like it. It stands out. I mean, it's pretty huge if this happens. I expect it to happen maybe 1 time ever TBH. (so much work for something so unlikely 🙄 )

cormacmccarthy:
image

PR #72, opened 2017-12-07
Gh41   more flavor text for tastier hp messaging

### LINKED ISSUE #41: More flavor text for tastier HP messaging
Right now, no matter what your HP is, it just says:

Baxter has NHP left.

There should be more flavorful messaging.

- [x] "Bloodied" for when you first get lower than 1/2 HP. "Baxter is bloodied, with only 15HP left."
- [x] "only" for when you are bloodied. "Baxter only has 3HP left."

cormacmccarthy:
fixes #41


game.js
```diff
@@ -302,11 +302,14 @@ ${monsterCard(monster, contestant.lastMonsterPlayed !== monster)}`
 			icon = '🏓';
 		}
 
+		const bloodied = (prevHp > monster.bloodiedValue) ? `${monster.givenName} is now bloodied. ` : '';
```
tweettypography: Shouldn't this be `prevHp >= monster.bloodiedValue`?

PR #80, opened 2017-12-11
Forked stick

cormacmccarthy:
Adds a "forked stick" card.

Easier to hit targets, does less damage to Minotaurs (trapping their head and stomping on it seems... not very effective to Minotaurs) but extra damage to Basilisks (because that's precisely who it is intended to be used against).


cards/forked-stick.js
```diff
@@ -0,0 +1,88 @@
...
+	}
+}
+
+ForkedStickCard.cardType = 'Forked Stick';
+ForkedStickCard.creatureType = BASILISK;
+ForkedStickCard.probability = 30;
+ForkedStickCard.description = `A simple weapon fashioned for ${ForkedStickCard.creatureType}-hunting.`;
+ForkedStickCard.cost = 6;
+ForkedStickCard.level = 1;
+ForkedStickCard.permittedClasses = [FIGHTER, MINOTAUR];
```
tweettypography: MINOTAUR isn't a class...

cards/forked-stick.js
```diff
@@ -0,0 +1,88 @@
...
+		return this.options.damageModifier;
+	}
+
+	get attackModifier () {
+		return this.options.attackModifier;
+	}
+
+	getDamageModifier (target) {
+		if (target.name === this.creatureType) {
+			return this.options.damageModifier;
```
tweettypography: Should just be `this.damageModifier` (not `.options`)

monsters/base.js
```diff
@@ -42,8 +42,9 @@ class BaseMonster extends BaseCreature {
 	canHold (object) {
 		const appropriateLevel = (!object.level || object.level <= this.level);
 		const appropriateClass = (!object.permittedClasses || object.permittedClasses.includes(this.class));
+		const appropriateType = (!object.permittedTypes || object.permittedTypes.includes(this.type));
 
-		return appropriateLevel && appropriateClass;
+		return appropriateLevel && (appropriateClass || appropriateType);
```
tweettypography: This should probably be `appropriateClass && appropriateType`

game.js
```diff
@@ -181,6 +182,15 @@ ${monsterCard(monster, contestant.lastMonsterPlayed !== monster)}`
 		});
 	}
 
+	announceNarration (className, monster, { narration }) {
+		const channel = this.publicChannel;
+
+		channel({
+			announce:
+`${narration}`
```
tweettypography: You put this on another line but both the open and close ticks are on this line too so I'm not sure the reason?

tweettypography: Ps there's a version of this messaging action in `master` now for another purpose

monsters/base.js
cormacmccarthy: I really wanted for you to only have to specify one or the other. The point was that this would make it easier to specify which monsters were appropriate for the card, not that it would be harder. For instance, this lets you say "only basilisks can use this card" _or_ "any barbarian can use this card" without having to double-specify "you have to be a basilisk, which is also a barbarian".

tweettypography: That's already true though - these default allowing you to hold it (that is, if you don't set classes it assumes you're allowed)

cormacmccarthy: If I understand the code correctly, the way I have it now, if you set `permittedClasses = []` and `permittedTypes = [BASILISK]` this will only let the basilisk hold it. I mean, it's not a huge difference between that and `permittedClasses = [BARBARIAN]` and `permittedTypes = [BASILISK]`, but that complicates the code a tiny bit more and makes it more brittle if we ever change BASILISK to be something other than a BARBARIAN.

tweettypography: Here's what I mean:
`appropriateClass = (!object.permittedClasses || object.permittedClasses.includes(this.class));`

So if you don't specify `permittedClasses` at all then `appropriateClass` will be true. Same goes for `appropriateType` - it will be true if you don't specify `permittedTypes`.

That's why you can use `&&` - if you specify a level, class, and type it should require all of them but in most cases you won't specify at least one of those.

Oh, and setting `permittedClasses` to `[]` won't work - you want to not set it all (or set it to null or undefined) because setting it to an empty array is basically saying that no classes are allowed to hold the card...

cards/constrict.js
```diff
@@ -0,0 +1,53 @@
...
+	get stats () {
+		return `${super.stats}
+Chance to immobilize opponent by coiling around them and then squeezing.`;
+	}
+
+ConstrictCard.cardType = 'Constrict';
+ConstrictCard.strongAgainstCreatureTypes = [GLADIATOR, MINOTAUR];
+ConstrictCard.probability = 30;
+ConstrictCard.description = `Your body _is_ the weapon`;
+ConstrictCard.permittedClasses = [];
```
tweettypography: Remove this line

cards/enthrall.js
```diff
@@ -0,0 +1,65 @@
...
+				.then(() => !target.dead));
+		});
+	}
+
+EnthrallCard.cardType = 'Enthrall';
+EnthrallCard.level = 2;
+EnthrallCard.strongAgainstCreatureTypes = [GLADIATOR, BASILISK];
+EnthrallCard.probability = 30;
+EnthrallCard.description = `You strut and preen. Your beauty overwhelmes and enthralls everyone, except yourself.`;
+EnthrallCard.permittedClasses = [WEEPING_ANGEL];
```
tweettypography: Did you mean this to be `permittedTypes`? Otherwise it should be `CLERIC` because `WEEPING_ANGEL` is not a type

cards/entrance.js
```diff
@@ -0,0 +1,53 @@
...
+		return `${super.stats}
+Chance to immobilize and damage your opponents.`;
+	}
+
+EntranceCard.cardType = 'Enthrall';
+EntranceCard.strongAgainstCreatureTypes = [GLADIATOR, BASILISK];
+EntranceCard.probability = 20;
+EntranceCard.level = 3;
+EntranceCard.description = `You strut and preen. Your painful beauty overwhelmes and entrances everyone, except yourself.`;
+EntranceCard.permittedClasses = [WEEPING_ANGEL];
```
tweettypography: Same here - this should be CLERIC or `permittedTypes`. Personally these seem like things that could be for all clerics and not just angels

cards/mesmerize.js
```diff
@@ -0,0 +1,51 @@
...
+	get stats () {
+		return `${super.stats}
+Chance to immobilize everyone with your beauty.`;
+	}
+
+MesmerizeCard.cardType = 'Mesmerize';
+MesmerizeCard.strongAgainstCreatureTypes = [GLADIATOR, BASILISK];
+MesmerizeCard.probability = 30;
+MesmerizeCard.description = `You strut and preen. Your beauty overwhelmes and mesmerizes everyone, including yourself.`;
+MesmerizeCard.permittedClasses = [WEEPING_ANGEL];
```
tweettypography: You know what I'm going to say here...

cards/mesmerize.js
```diff
@@ -0,0 +1,51 @@
+/* eslint-disable max-len */
+
+const ImmobilizeCard = require('./immoblize');
+
+const { FIGHTER, BARBARIAN } = require('../helpers/classes');
+const { GLADIATOR, MINOTAUR, BASILISK } = require('../helpers/creature-types');
+const { DEFENSE_PHASE } = require('../helpers/phases');
+const { roll } = require('../helpers/chance');
+
+class MesmerizeCard extends ImmobilizeCard {
```
tweettypography: This is a base for enthrall and entrance but you redefine so many default values and properties in those cards - maybe you don't need to set `permittedTypes` etc again in the other cards?

cards/mesmerize.js
```diff
@@ -0,0 +1,51 @@
...
+		return `${super.stats}
+Chance to immobilize everyone with your beauty.`;
+	}
+
+MesmerizeCard.cardType = 'Mesmerize';
+MesmerizeCard.strongAgainstCreatureTypes = [GLADIATOR, BASILISK];
+MesmerizeCard.probability = 30;
+MesmerizeCard.description = `You strut and preen. Your beauty overwhelmes and mesmerizes everyone, including yourself.`;
+MesmerizeCard.permittedClasses = [WEEPING_ANGEL];
+MesmerizeCard.weakAgainstCreatureTypes = [MINOTAUR, WEEPING_ANGEL];
```
tweettypography: Minotaurs have strong willpower? Makes sense to me

cards/mesmerize.js
cormacmccarthy: Fixed.

cards/mesmerize.js
cormacmccarthy: Haha, actually what I was thinking is that they would be less likely to find the angel beautiful. They'd find them about as beautiful as we'd find a cow beautiful. So, this attack is weak against them, but _could_ work. After all we have entire fairs where people stand around to oggle magnificent bovine.

cards/entrance.js
cormacmccarthy: I suppose that's true... We don't have any other cleric cards yet though and this form of immobilization seemed particularly unique to Angels. If we had other clerics I was thinking of them using something more pedestrian, like "paralyze", or maybe some sort of fear based immobilization instead of beauty based.

tweettypography:
@cormacmccarthy This looks good to me now, just needs some unit tests and it'll be ready to merge

cards/constrict.js
```diff
@@ -0,0 +1,49 @@
...
+	doDamageOnImmobilize: true,
+	freedomThresholdModifier: 3
+};
+ConstrictCard.action = ['constrict', 'constricts', 'constricted'];
+
+ConstrictCard.flavors = {
+	hits: [
+		[`${ConstrictCard.action[1]}`, 80],
+		['squeezes and squeezes', 50],
+		['tightens so hard that anything on the inside that could easily come to the outside, well... _does_', 5]
```
tweettypography: This doesn't really make sense because it's `player 1 <description> player 2`

cards/enthrall.js
```diff
@@ -0,0 +1,71 @@
...
+	attackModifier: 2,
+	hitOnFail: false,
+	freedomThresholdModifier: 1
+};
+
+EnthrallCard.flavors = {
+	hits: [
+		['You enthrall your adversaries', 80],
+		['Your natural beauty overwhelms your enemies', 30],
+		['Narcisus himself would be distracted by your beauty... and that\'s when you hit.', 5]
```
tweettypography: None of these actual make sense - unless you changed the way these are used?

tweettypography: *Narcissus

cards/forked-stick.js
```diff
@@ -0,0 +1,48 @@
...
+ForkedStickCard.permittedClassesAndTypes = [FIGHTER, BARBARIAN];
+ForkedStickCard.weakAgainstCreatureTypes = [MINOTAUR];
+ForkedStickCard.defaults = {
+	...ImmobilizeCard.defaults
+};
+ForkedStickCard.action = ['pin', 'pins', 'pinned'];
+
+ForkedStickCard.flavors = {
+	hits: [
+		[`${ForkedStickCard.action[1]} head to the ground`, 80],
```
tweettypography: "Baxter pins head to the ground Legolost?" That doesn't make sense...

cards/immobilize.js
```diff
@@ -0,0 +1,227 @@
...
+								reason: `and needs higher than a ${this.getFreedomThreshold(player, target)} to be free`,
+								card: this,
+								roll: freedomRoll,
+								player: target,
+								target: player,
+								outcome: success ? commentary || `Success! ${target.givenName} is freed.` : commentary || `${target.givenName} remains ${this.getAction(2)} and will miss a turn.`
+							});
+
+							if (success) {
+								target.encounterEffects = target.encounterEffects.filter(effect => effect.effectType !== 'ImmobilizeEffect');
```
tweettypography: Okay, I wrote this logic originally but I just realized it has an issue - you'll remove all immobilization effects. If a different person has also immobilized you (or if the same person laid this card twice) it shouldn't get cleared. One idea might be to have the card generate a random id for itself when it is instantiated which can be used for this.

tweettypography:
Okay just a few comments, mainly about flavor text. Otherwise looks good

cards/immobilize.js
iowillhoit: But you wouldn't make it here if you were already immobilized. [line](https://github.com/deck-monsters/deck-monsters/pull/80/files#diff-fcfc4784967c1db04ef6552f11dc2aeaR94)

cormacmccarthy: That's true, someone can't be immobilized twice right now

I could see an argument that two people should be able to play forked stick against one person and double-immobilize them, but for now it doesn't work like that.

cards/constrict.js
cormacmccarthy: Good call. I forgot to ever come back to this... I'll take another pass at it now.

cards/immobilize.js
tweettypography: Ah, good point - so if someone has already pinned you then another person can't _also_ pin you. That makes sense. It doesn't come into play too much for regular sticks anyway because of the way targeting currently works

cards/entrance.js
```diff
@@ -0,0 +1,49 @@
...
+	doDamageOnImmobilize: true,
+	freedomThresholdModifier: 2
+};
+EntranceCard.actions = ['entrance', 'entrances', 'entranced'];
+
+EntranceCard.flavors = {
+	hits: [
+		[EntranceCard.actions[1], 80],
+		['uses their painfully stunning natural beauty to bring falling to their knees in agonized worship', 30],
+		[`${EntranceCard.actions[1]} even Narcissus himself with their beauty... And that's when they sucker punche`, 5]
```
tweettypography: Did you mean this to be `punche`?

cards/forked-stick.js
```diff
@@ -0,0 +1,48 @@
...
+ForkedStickCard.defaults = {
+	...ImmobilizeCard.defaults
+};
+ForkedStickCard.actions = ['pin', 'pins', 'pinned'];
+
+ForkedStickCard.flavors = {
+	hits: [
+		[`${ForkedStickCard.actions[1]} to the ground, the head of`, 80],
+		[`${ForkedStickCard.actions[1]} to the wall, the neck of`, 50],
+		['in a fit of brute strength, snags by the neck, and brutally lofts into the air, where they dangle like a toddler\'s booger', 5]
```
tweettypography: This flavor test still doesn't fit the pattern...

cards/index.js
```diff
@@ -83,9 +95,9 @@ const fillDeck = (deck, options, creature) => {
 const getInitialDeck = (options, creature) => {
 	// See above re: options
 	const deck = [
-		new HitCard(),
-		new HitCard(),
-		new HitCard(),
+		new MesmerizeCard(),
+		new ForkedStickCard(),
+		new ConstrictCard(),
```
tweettypography: I'm assuming this was for testing? We should put it back

PR #81, opened 2017-12-12
Take damage modifiers into account with max double damage for nat 20

cormacmccarthy:
image

I should have done 15 damage, not 12...


RodEsp:
What? This is just going to make crits more ridiculous than they already are!

tweettypography:
@RodEsp agreed, it's going to get nuts haha

cormacmccarthy:
I agree. What really needed changed here was the messaging. I believe David already changed/fixed this.

PR #88, opened 2017-12-14
Look at monsters in detail command

iowillhoit:
I use `look at monsters` to see if my monsters have enough health to go back into the ring. With the monster descriptions I have to scroll up to look at all my monsters (especially when I'm on my phone). 

This changes `look at monsters` to hide the description, and adds the command `look at monsters in detail` to see the description.


tweettypography:
Want to add a quick unit test to this code since you're testing it?

PR #90, opened 2017-12-14
Look at monster in detail.

iowillhoit:
Refactors and adds unit tests

### DIFF
--- a/game.js
+++ b/game.js
@@ -570,12 +570,8 @@ ${monsterCard(monster)}`
 					return character.lookAtCards(channel, deckName)
 						.catch(err => log(err));
 				},
-				lookAtMonsters () {
-					return character.lookAtMonsters(channel, false)
-						.catch(err => log(err));
-				},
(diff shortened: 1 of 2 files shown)


iowillhoit:
@tweettypography let me know if you can think of a better way to test this...

game.js
```diff
@@ -570,12 +570,8 @@ ${monsterCard(monster)}`
 					return character.lookAtCards(channel, deckName)
 						.catch(err => log(err));
 				},
-				lookAtMonsters () {
-					return character.lookAtMonsters(channel, false)
-						.catch(err => log(err));
-				},
-				lookAtMonstersInDetail () {
-					return character.lookAtMonsters(channel, true)
+				lookAtMonsters ({ inDetail }) {
```
tweettypography: Only thing I would say is this should probably have a default of an empty object: `lookAtMonsters ({ inDetail } = {})`

PR #101, opened 2017-12-19
Assign XP dynamically based on monster level difference

### LINKED ISSUE #87: XP should be relative based on level
Something like:

winner level > loser level = XP-(winner level-loser level)
winner level == loser level = XP
winner level < loser level = XP+(loser level-winner level)

RodEsp:
Fixes #87 

This calculates XP gained by monsters using a logarithmic funciton as described in #87.

With this new logic player characters will still gain 10 XP per victory but monsters will only gain XP when they deal the killing blow to another monster. Draws will award no XP to the monster.

The breakdown essentially goes like this:
```
|-----------------------------|
| Level Difference	| XP Gain |
|-----------------------------|
|		0			|	10	  |
|		1			|	5	  |
|		2			|	3	  |
|		3			|	1	  |
|		4			|	1	  |
|		5+			|	0	  |
|-----------------------------|
```

If a monster kills a higher level monster than itself the breakdown is the following:
```
|-----------------------------|
| Level Difference	| XP Gain |
|-----------------------------|
|		0			|	10	  |
|		1			|	20	  |
|		2			|	40	  |
|-----------------------------|
```
And it exponentially grows from there.


ring/index.js
```diff
@@ -508,6 +516,21 @@ class Ring extends BaseClass {
...
 	}
 
+	awardMonsterXP ({ contestants }) {
+		contestants.forEach((contestant) => {
+			const { monster } = contestant;
+			const monsterXP = calculateXP(monster, contestants);
+			
+			monster.xp += monsterXP;
+
+			this.emit('gainedXP', {
```
tweettypography: We should probably only emit this if we actually gained xp, right? So if the value of monsterXP is 0 we can skip

tweettypography: Also it'd probably be fairly easy to unit test this function - seems like a good idea to get that in here so we don't keep adding code without tests

helpers/experience.js
```diff
@@ -0,0 +1,27 @@
+const XP_PER_VICTORY = 10;
+const XP_PER_DEFEAT = 1;
+const BASE_XP_PER_KILL = 10;
+const STARTING_XP = 0;
+
+// levels are a Fibonacci sequence (100, 200, 300, 500, 800, 1300, 2100, ...)
```
tweettypography: This comment should be changed or removed

creatures/base.js
```diff
@@ -264,6 +265,22 @@ Battles won: ${this.battles.wins}`;
...
+		return this.killedCreatures;
+	}
+
+	set killed (creature) {
+		if (!this.killedCreatures) {
+			this.killedCreatures = [creature];
+		} else {
+			this.killedCreatures.push(creature);
+		}
+	}
```
tweettypography: I don't see you clearing `this.killedCreatures` anywhere which means it will just continue to grow and grow over time. If you set it as a property of the encounter then it will be automatically cleared. See https://github.com/deck-monsters/deck-monsters/blob/master/creatures/base.js#L221-L230 for example, which gets cleared here: https://github.com/deck-monsters/deck-monsters/blob/master/creatures/base.js#L437

The only downside is you'd have to save the value somewhere before you ended the encounter because you don't do anything with it until _after_ the encounter has ended. Maybe just copy it from the `monster` to the `contestant` before calling `contestant.monster.endEncounter();`? I'll put an example on that code.

ring/index.js
```diff
@@ -449,6 +450,11 @@ class Ring extends BaseClass {
 			}
 		});
 
+		// End the encounter for all monsters
+		contestants.forEach((contestant) => {
+			contestant.monster.endEncounter();
+		});
```
tweettypography: Could be:

```javascript
contestants.forEach((contestant) => {
  contestant.killed = contestant.monster.killed;
  contestant.monster.endEncounter();
});
```

ring/index.js
RodEsp: Hmm, I'm not sure how to test just that function without running a full simulation and proxyquiring calculateXP to ensure it was called. Suggestions?

tweettypography: I would have generated an array of fake contestants with different levels - at least three - and artificially assigned them to each other's `killed` slots. Then you could test that the change in xp from before and after running this function made sense.

PR #106, opened 2017-12-20
Feature/look at monster

iowillhoit:
If you own the monster, `look at monster <name>` will show you it's entire hand


characters/beastmaster.js
```diff
@@ -1,4 +1,5 @@
 const moment = require('moment');
+const { some } = require('lodash');
```
tweettypography: I don't think `lodash` by itself is one of our dependencies

PR #107, opened 2017-12-20
(GH82) announce card drop in ring

### LINKED ISSUE #82: Card drops as a result of a fight should be public

cormacmccarthy:
fixes #82

### DIFF
--- a/game.js
+++ b/game.js
@@ -137,16 +137,23 @@ class Game extends BaseClass {
 	}
 
 	announceCardDrop (className, game, { contestant, card }) {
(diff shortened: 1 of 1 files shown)


game.js
```diff
@@ -137,17 +137,24 @@ class Game extends BaseClass {
...
+${cardDropped}`;
 
-${cardDropped}`,
+		this.channelManager.queueMessage({
+			announce,
 			channel,
 			channelName
 		});
+
+		channel({
```
RodEsp: Don't you mean `publicChannel` here?

PR #111, opened 2017-12-22
Separate out announcements

cormacmccarthy:
`game.js` was growing very large. Separating out announcements will make it easier to add new features and to unit test announcements.


tweettypography:
@cormacmccarthy When you get a chance can you update this to match the latest game changes? I didn't get a chance to review it right away as it was a bigger change but it looks good if conflicts are resolved

PR #113, opened 2017-12-24
fixes flee card mechanics

cormacmccarthy:


tweettypography:
I don’t think this should be against AC. How about a standard 10 every time for now?

PR #124, opened 2017-12-27
Grant losers xp

cormacmccarthy:


helpers/experience.js
```diff
@@ -1,25 +1,51 @@
...
+		gainedXP += xpFormula(levelDifference, BASE_XP_PER_KILL);
 	});
 
+	if (contestant.killedBy) {
+		levelDifference = monster.level - contestant.killedBy.level;
+
+		gainedXP += xpFormula(levelDifference, BASE_XP_PER_KILLED_BY);
+	} else {
+		// XP for being the last monster standing or fleeing
+		const averageLevelDifference = monster.level - getAverageLevel(contestants);
```
tweettypography: So it looks like you calculate the average of all levels including your own, which means the results will be a little skewed. For example, if you are a level 10 and you fight two level 1s you'll get an average of 4 which makes the difference 6 when it should be 9. It'd be best to filter yourself out of the collection.

ring/index.js
```diff
@@ -506,6 +505,8 @@ class Ring extends BaseClass {
 			}
 		});
 
+		contestants.forEach(contestant => this.awardMonsterXP(contestant, contestants));
```
tweettypography: Why do we loop again here? Couldn't you just do this as part of the loop above (outside of the conditional)?

helpers/experience.mocha.node.js
```diff
@@ -1,40 +1,288 @@
...
+					fled: true
+				};
+
+				const contestants = [constestant1, constestant2];
+
+				expect(calculateXP(constestant1, contestants)).to.equal(48);
+				expect(calculateXP(constestant2, contestants)).to.equal(0);
+			});
+
+			it('assigns 32 XP if you flee a monster 4 levels higher', () => {
```
tweettypography: @cormacmccarthy This is kind of crazy - we should probably limit the amount you can earn from fleeing to 5. Otherwise you could send you monster into battles with stronger monsters all the time and fill your hand with flee cards.

ring/index.js
cormacmccarthy: At one point you couldn't because this part needed to be populated for all contestants:

```
			contestant.killed = contestant.monster.killed;
			contestant.killedBy = contestant.monster.killedBy;
			contestant.fled = contestant.monster.fled;
			contestant.rounds = rounds;
			contestant.encounter = contestant.monster.endEncounter();
```

But I'll re-confirm because tons has changed since I broke this part out of the loop...

helpers/experience.mocha.node.js
cormacmccarthy: Agreed. I'll set it as max 5 times the number of rounds fought.

tweettypography: Makes sense to me, though maybe we should track and use the number of rounds the monster was in (before fleeing or dying) rather than the total rounds fought

cormacmccarthy:
Still not ready, but getting closer. I think just unit tests need filled in at this point.

cormacmccarthy:
ready.

game.js
```diff
@@ -126,11 +127,13 @@ class Game extends BaseClass {
 
 		let killedMessage = '';
 		if (killed) {
-			killedMessage = ` for killing ${killed.length} ${killed.length > 1 ? 'monsters' : 'monster'}.`;
+			killedMessage = ` for killing ${killed.length} ${(killed.length > 1 || killed.length === 0) ? 'monsters' : 'monster'}.`;
 		}
 
+		const reasonsMessage = (reasons) ? `\n\n${reasons}` : '';
```
tweettypography: Putting \n in a template string will simply output \n

helpers/experience.js
```diff
@@ -1,27 +1,77 @@
...
 
-		// This formula calculates XP based on the following logarithmic function
-		// -0.3x = log(y) where x is the level difference of the two monsters
-		gainedXP += Math.round(Math.pow(10, (-0.3 * levelDifference)) * BASE_XP_PER_KILL);
+		gainedXP += xp;
 	});
 
-	return gainedXP;
+	if (contestant.killedBy) {
+		levelDifference = monster.level - contestant.killedBy.level;
```
tweettypography: Maybe we should say “if level difference <0, skip this” (don’t award XP when a lower level monster kills you)

helpers/experience.js
```diff
@@ -1,27 +1,77 @@
...
+		xp = xpFormula(levelDifference, BASE_XP_PER_KILLED_BY);
+
+		reasons.push(`Gained ${xp} XP for being killed by ${contestant.killedBy.givenName} (${levelDifference * -1} level difference)`);
```
tweettypography: You need a special case for when the level difference is 0.

helpers/experience.js
```diff
@@ -1,27 +1,77 @@
...
+
+		gainedXP += xp;
+	} else {
+		// XP for being the last monster standing or fleeing
+		const avgLevel = getAverageLevel(monster, contestants);
+		const averageLevelDifference = monster.level - getAverageLevel(monster, contestants);
```
tweettypography: In this scenario I think it’s fine if the level is less than or equal to 0

helpers/experience.mocha.node.js
```diff
@@ -1,40 +1,876 @@
...
+			({ gainedXP, reasons } = calculateXP(constestant3, contestants));
+			expect(gainedXP).to.equal(11);
+			expect(reasons).to.equal('Gained 10 XP for killing fred (0 level difference)\nGained 1 XP for being killed by sonka (0 level difference)');
+			({ gainedXP } = calculateXP(constestant4, contestants));
+			expect(gainedXP).to.equal(11);
+			({ gainedXP } = calculateXP(constestant5, contestants));
+			expect(gainedXP).to.equal(13);
+		});
+
+		it('assigns proper xp to winners and losers if one beats all all different levels', () => {
```
tweettypography: Just fix the test description here

helpers/experience.js
cormacmccarthy: I think I'd be ok with if a monster only 1 level lower than you kills you its fine, but 2 or more, don't award XP.

cormacmccarthy: You only get 1 xp if a monster 1 level lower than you kills you... If they are 2 or more below you, you get nothing. Add a unit test to show/ensure this.

helpers/experience.js
```diff
@@ -1,27 +1,78 @@
...
+
+		gainedXP += xp;
+	} else {
+		// XP for being the last monster standing or fleeing
+		const avgLevel = getAverageLevel(monster, contestants);
+		const averageLevelDifference = monster.level - avgLevel;
+		const xpBase = contestant.fled ? BASE_XP_PER_FLEEING : BASE_XP_LAST_ONE_STANDING;
+		xp = Math.min(xpFormula(averageLevelDifference, xpBase), 5 * rounds);
+
+		reasons.push(`Gained ${xp} XP for ${contestant.fled ? 'fleeing' : 'last one standing'} in battle with opponents with average level of ${avgLevel}`);// eslint-disable-line max-len
```
tweettypography: Maybe “being the last one standing”?

PR #127, opened 2017-12-31
tweaks xp related text a bit

cormacmccarthy:


cormacmccarthy:
decided to use Math.round instead of Math.ceil so that you wouldn't have a fight with all beginner level monsters getting credited as if they were all level 1s.

PR #128, opened 2018-01-01
Gh126   one special level 0 rare card per class  each player starts with one in their pool

### LINKED ISSUE #126: One special level 0 rare card per class, each player starts with one in their pool.
We need a few other level 0 special cards that are specific to class:

- [x] Coil for basilisk (a weaker form of constrict)
- [x] Horn Gore for Minotaur (2 chances for 1d4 hits)
- [x] Berserk for gladiator (punch for 1 damage and keep punching until you miss)
- [x] Blink for Weeping Angel (your opponent is effectively taken out of the battle flow for 1 turn --they can not be targeted or play cards during that time. On the turn they lost, if you are still alive, you drain 1d4 hp and 4d4 xp from them). 

These should be rare cards but each person should either start with one of each (so that each of their monsters will have a single special Big Card) or at the least 1 randomly seeded in their pool.

cormacmccarthy:
closes #126 
 
Adds coil, horn gore, berserk, and blink. A special super rare level 0 card unique to each monster type.


tweettypography:
Going to take me a bit longer to review this. Can you make sure that conflicts are resolved first?

cormacmccarthy:
@tweettypography merged master

cards/berserk.js
```diff
@@ -0,0 +1,106 @@
...
+};
+
+BerserkCard.flavors = {
+	hits: [
+		['punches', 80],
+		['smacks', 70],
+		['back-fists', 70],
+		['upper-cuts', 70],
+		['head-butts', 50],
+		['puts a boot to the face', 50],
```
RodEsp: Should this be `puts a boot to the face of`?

cards/berserk.js
```diff
@@ -0,0 +1,106 @@
...
+		});
+	}
+}
+
+BerserkCard.cardType = 'Berserk';
+BerserkCard.probability = 20;
+BerserkCard.description = 'The whole world disappears into a beautiful still, silent, red. At the center of all things is the perfect face of your enemy. Destroy it.';
+BerserkCard.cost = 6;
+BerserkCard.level = 0;
+BerserkCard.permittedClassesAndTypes = [GLADIATOR];
```
RodEsp: I feel like, thematically, this should be usable by Barbarians.

cards/blink.js
```diff
@@ -0,0 +1,196 @@
...
+					card,
+					phase
+				}) => {
+					const { effect } = card;
+
+					if (effect && phase === DEFENSE_PHASE) {
+						card.effect = (player, target, effectRing, effectActiveContestants) => {
+							if (target === blinkTarget) {
+								this.emit('effect', {
+									effectResult: `not target-able because they are ${this.icon}  time-shifted by`,
```
RodEsp: extra space here?

cards/blink.js
```diff
@@ -0,0 +1,196 @@
...
+								return Promise.resolve(true);
+							}
+							return effect.call(card, player, target, effectRing, effectActiveContestants);
+						};
+					} else if (phase === ATTACK_PHASE) {
+						const turnsLeftToBlink = this.turnsToBlink - blinkTarget.blinkedTurns;
+						if (turnsLeftToBlink && !blinkPlayer.dead) {
+							blinkTarget.blinkedTurns++;
+
+							const effectResult = `${this.icon}  time-shifted for ${turnsLeftToBlink} more turn${turnsLeftToBlink > 1 ? 's' : ''} by`;
```
RodEsp: also extra space here?

cards/blink.js
```diff
@@ -0,0 +1,196 @@
+/* eslint-disable max-len */
```
RodEsp: Could we just call this timeshift instead of blink?

cards/berserk.js
cormacmccarthy: Thematically, I agree. The intention here is that we are giving a pretty powerful card to each type that they can use out of the gate. So if we gave it to Barbarians, then the Minotaur would have Berserk _and_ Horn Gore to use starting off instead of just Horn Gore. Can we pretend like this is somehow only able to be done by humans at this level? Or we could make this a level 1 card usable by Barbarians and think up something else for level 0 Gladiators...

cormacmccarthy: Actually, turns out that Gladiator isn't a barbarian at all. It's a fighter. I think I'll make this card a level 1 card for Barbarians and clone it and make the clone a level 0 card usable by fighters.

cards/blink.js
cormacmccarthy: I believe we typically put two spaces following icons because otherwise the formatting can be weird.

tweettypography:
Do you think there should be a limit on battle focus / berserk? I could see a high-level monster using it and hitting 10, 20, or even 100 times, especially against low-level monsters.

cormacmccarthy:
The part of me that used to play Killer Instinct says that it should be unlimited and that we should add announcements for Ultra and Ultimate combos (or, if you're a MK guy, "fatalities" I guess was their equivalent?).

I think maybe we should just see what happens and see if it ends up just being completely OP. It is true that you could totally end up perma-killing your opponent with this, especially if they are low level, but I think that that maybe adds more incentive not to just stupidly send a low level opponent against a high level one.

cormacmccarthy:
https://www.youtube.com/watch?v=xqCsB5B8ZcY

tweettypography:
Fair enough - and I guess we do have rolls of 1 as a bit of a safeguard against an infinite loop, but let's say you send in a monster and someone softens you to an AC below their attack modifier - then they'll hit _every time_ until they roll a 1, meaning that even high-level monsters are likely to be perma-deathed by this

Issue #11, opened 2017-05-15
Card pops


cormacmccarthy:
When you roll a natural 20 or stroke of luck with a card, it should level up to the next version of itself.

ie:
Hit -> Hit Harder -> Pound
Heal -> Revive

Should this happen mid-battle, or after the battle is over? If mid-battle it will screw up your energy calculations you built your deck around.

tweettypography:
A version of this exists in a single card now. Not sure we want to implement this across the board - we can potentially reopen later if we decide that'd be a good idea. Best solution is probably to make more cards that have this as a specific part of that card.

cormacmccarthy:
New version of this idea:

There's a magical device you can put your card in and then feed coins. Every time you feed it a coin, there's a chance (maybe 1:20) it will pop the card, but there's _also_ a chance the card will be incinerated (maybe 1:100). It's more interesting/fun than just trading coins for card upgrades or selling old cards and getting new cards.

In DM:
`jane send hit to the popper`
`how many coins would you like to send? (you have 35)`
`10`
`sent hit to the popper with 10 coins`

In #deck-monsters-popper
`Christopher McCulloh sent hit to the popper with 10 coins`
`The popper begins to glow with the warm comforting light of radioactive energies as the coins fall in one by on`
`GLOP - The card wiggles and fuzzes a bit as a coin disintegrates.`
`GLOP - The card [flavor] as a coin disintegrates.`
`GLOP - The card [flavor] as a coin disintegrates.`
`POP - A trans dimensional crack rings out as the card seems to simultaneously ex and im plode and suddenly, gloriously, mysteriously, hit harder is now laying in the popping chamber`
`hit was transmogrified into hit harder for Christopher McCulloh after 5 coins went into the popper`

Back in DM:
`hit harder returns from the popper with 6 coins. hit and 4 coins are lost forever.`

PR #130, opened 2018-01-04
Gh11   card pops

cormacmccarthy:


cormacmccarthy:
Going in a new direction. Updated #11 with new requirements.

PR #139, opened 2018-01-05
Gh135   fix pin logic

### LINKED ISSUE #135: Fix pin logic

cormacmccarthy:
fixes #135 

Makes it easier for pinned player to break free the longer they are pinned.


cards/immobilize.js
```diff
@@ -116,6 +121,7 @@ class ImmobilizeCard extends HitCard {
 				});
 
 				if (attackSuccess.success) {
+					target.pinnedTurns = 0;
```
tweettypography: I know, this gets reset, but I'm not crazy about editing the actual player here as you'd be editing their master instance. I don't see that you made `pinnedTurns` part of their encounter properties, so they wouldn't get cleaned up. An alternative would be tracking it on the card (as that is a copy) or adding it to encounter properties.

cormacmccarthy: I'll figure out how to add to encounter properties... Pretty sure there's an example of it somewhere.

cormacmccarthy: @tweettypography done.

PR #141, opened 2018-01-05
Adds a remove card method to beastmaster

cormacmccarthy:


characters/base.js
```diff
@@ -47,6 +47,19 @@ class BaseCharacter extends BaseCreature {
 		this.emit('cardAdded', { card });
 	}
 
+	removeCard (cardToRemove) {
+		let alreadyRemoved = false;
+		this.deck = this.deck.filter((card) => {
+			const keep = alreadyRemoved || cardToRemove.name !== card.name;
```
tweettypography: If you only compare on `name` then you don't take into account cards that have been modified (like the Kalevala). See https://github.com/deck-monsters/deck-monsters/blob/master/cards/index.js#L143 for a slower but more comprehensive compare.

cormacmccarthy:
@tweettypography fixed

PR #147, opened 2018-01-06
(GH146) allows for custom hit icons

### LINKED ISSUE #146: Hit announcement should allow for custom icons and thresholds
Not every hit looks the same. In addition to the custom flavors we already have, we should allow for custom icons.

cormacmccarthy:
fixes #146


cormacmccarthy:
This will come into play a lot more in exploration, but will also be important for constrict whenever we get around to revisiting flavor text there.

helpers/flavor.js
```diff
@@ -7,7 +7,7 @@ const flavor = {
 
 		const words = possibleWords.find(word => percent() <= word[1]) || flavor.getFlavor(category, flavors);
 
-		return words[0];
+		return words;
```
tweettypography: Not sure I really understand the implications of this off the top of my head. Why were we previously getting the first element in an array? Did we really update every place we would need to make that that irrelevant?

cards/base.js
```diff
@@ -44,6 +44,42 @@ class BaseCard extends BaseClass {
...
+	get icons () {
+		return this.options.icons;
+	}
+
+	set icons (icons) {
+		this.setOptions({
+			icons
+		});
+	}
+
```
tweettypography: You're setting (well, preparing to set) a lot of things to the options right now. Things we set on the options are stored in the game state, so this has the potential to really blow up our saved state (that is, to make it really big). What are you using these for and how frequently? Are they really state-dependent or could they be determined programmatically (and therefore not need to be in state).

announcements/hit.mocha.node.js
```diff
@@ -163,5 +163,137 @@ describe('./announcements/hit.js', () => {
...
+`;
+			const publicChannel = ({ announce }) => {
+				expect(announce).to.equal(announcement);
+			};
+
+			const monster = new Gladiator({ name: 'monster', hpVariance: 0, acVariance: 0 });
+			const assailant = new Gladiator({ name: 'assailant', hpVariance: 0, acVariance: 0 });
+			const card = {
+				flavorText: '🔥 💪  Monster is caught in a sudden fire and takes 1 damage.'
+			};
```
tweettypography: I don't really understand how this would work. You'd set the flavor text on the card itself? Rather than including it as a parameter of the event?

announcements/hit.js
```diff
@@ -1,28 +1,40 @@
...
+		const icons = (card && card.flavorIcons) || defaultIcons;
+		icons.sort((a, b) => b.floor - a.floor);
+		({ icon } = icons.find(i => damage >= i.floor));
 	}
 
 	const bloodied = (monster.bloodied && prevHp > monster.bloodiedValue) ? `${monster.givenName} is now bloodied. ` : '';
 	const only = (monster.bloodied && monster.hp > 0) ? 'only ' : '';
+	const defaultFlavorText = `${assailant.icon} ${icon} ${monster.icon}  ${assailant.givenName} ${flavor[0]} ${monster.givenName} for ${damage} damage.`;
+
+	const flavorText = (card && card.flavorText) || defaultFlavorText;
```
tweettypography: Okay, not sure about getting this from the card, but if you're going to do this it'd be better to check for a "premade" flavor _first_ and then calculate the default if you need it. Right now you're doing all that work for nothing.

cards/base.js
tweettypography: To me it looks like you're using these two set temporary values on cards that you pass to the events. Feels like it might make more sense to pass messages as part of the event itself instead, but if it's easier to do it through the card (I can see that passing the message through might be messy and you're already passing the card) then you still don't need to set it to the cards options. Just set a property directly on the card. That is, get rid of the `flavorText ()` getter and setter and simply write `this.flavorText = 'text'` and `text = card.flavorText` as needed. There's nothing wrong with regular old properties if you don't need them saved to the game state. If you have a specific card that has hardcoded flavor text you could always include a getter that pulls the text from the constructor in it instead (but no setter required). Not sure about the flavor / icons props but you get the idea.

cormacmccarthy:
See #155 for examples of how this is going to be used. Specifically the hazard card: https://github.com/deck-monsters/deck-monsters/pull/155/files#diff-457783f08d8f4cbba2dda6365a9a40bfR84

helpers/flavor.js
cormacmccarthy: We were just getting back the actual flavor text and didn't care about the probability. But I wanted to send back both the text and the icon (if there was one) and the places where we were using this were surprisingly few so I moved the `[0]` accessor to there

I could have easily returned `{ text: words[0], icon: words[2] }` and then accessed the return value like `flavor.text` or `flavor.icon`, but... I didn't. Maybe I will though. I kinda like that better.

announcements/hit.mocha.node.js
cormacmccarthy: This announcement is triggered by the player object:

```
	hit (damage = 0, assailant, card) {
		if (this.hp < 1) return false;

		const hp = this.hp - damage;
		const originalHP = this.hp;

		this.hp = hp;

		this.emit('hit', {
			assailant,
			card,
			damage,
			hp,
			prevHp: originalHP
		});

		if (hp <= 0) {
			return this.die(assailant);
		}

		return true;
	}
```

Flavor stuff is already part of the card. This allows the _card_ to choose the specific flavor shown if necessary instead of the announcement just picking one semi-randomly.

cormacmccarthy: This is particularly important for exploration where there are a lot of different scenarios for hazards and the icon + text needs to match each other (otherwise you end up with really weird stuff like `🔥 🌟 LegoLost fell down a hole for 4 damage`

cormacmccarthy:
Ready for you now @tweettypography

cards/base.js
```diff
@@ -44,6 +44,12 @@ class BaseCard extends BaseClass {
 		return this.constructor.flavors || this.options.flavors;
 	}
 
+	set flavors (flavors) {
+		this.setOptions({
+			flavors
+		});
+	}
```
tweettypography: What is this for?

cormacmccarthy:
Fixed.

PR #155, opened 2018-01-09
Basic exploration framework

### LINKED ISSUE #12: Exploration
Should be able to send monsters out exploring.

For MVP what I think I'll have happen is, you send your monster out exploring `jane send baxter exploring`. Your monster is gone for 1 hour. They will have a chance of finding a card. They will have gotten into fights (which for MVP we won't see) and will loose a random number up to all but 1HP and have a chance of getting either lose or win XP.

Eventually I'm imagining them going out exploring and randomly getting into fights (which you may or may not get to watch) and if they die, you have to send another monster out to find them or else just dismiss them when they don't return. When they explore they have a chance at finding cards, but also have a chance at finding dungeons. Once your monster locates a dungeon you'll be able to specifically send them to that dungeon. The dungeons will be the kind of thing where you specifically tell the monster where to go, but you have to tell them ahead of time. You'll have an exploration deck that you build and try and the monster goes and executes your commands and then comes back in an hour and reports what happened.

The exploration deck would be simple and you'd have an infinite number of cards maybe. Basically would just consist of "left", "right", "straight", "search", and "random". If they encounter monsters they fight them. If they die, they're dead and you have to send another monster to get them. When they get to the end of the deck you give them, they just come home and tell you what they found.

This could look like this:
```
jane explore [dungeon] with straight, left, straight, search, right, search

[monster] enters [dungeon]
[monster] sees before [pronoun] a [dungeon theme adjectives] hallway
⬆️ straight
[monster] moves forward down the [dungeon theme adjectives] hallway
[monster] sees before [pronoun] a [dungeon theme adjectives] hallway
💢 [monster] encounters a [size][creature]
[monster] quickly dispatches [size][creature], losing [1d4] health but gaining [1d4] xp.
⬅️ left
[monster] can not go left
[monster] sees before [pronoun] a [dungeon theme adjectives] hallway
⬆️ straight
[monster] moves forward down the [dungeon theme adjectives] hallway
[monster] enters a [dungeon theme adjectives] ante-chamber. There are two doors: right & left.
🔄 search
[monster] searches the [dungeon theme adjectives] ante-chamber. Finds a door new door!
[monster] enters a [dungeon theme adjectives] ante-chamber. There are three doors: right, straight, & left.
➡️ right
[monster] enters the door on the right
[monster] sees before [pronoun] a [dungeon theme adjectives] hallway
🔄 search
[monster] searches the [dungeon theme adjectives] hallway. Finds a card!
[monster] runs out of cards and comes home with the fruits of [pronoun] labor.
```

cormacmccarthy:
fixes #12 

Adds a basic framework for Exploration with Nothing, Hazard, and Death discoveries.


exploration/discoveries/hazard.js
```diff
@@ -0,0 +1,99 @@
...
+				[`finds ${player.pronouns[1]}self caught in briars and takes`, 50, '🌵'],
+				['walks through a patch of stinging nettles and takes', 50, '🌿'],
+				['is eaten up by mosquitos for', 60, '🐜'],
+				['is bitten by a large spider for', 40, '🕷'],
+				['eats some mildly rotten trail rations and falls ill for', 5, '🤢'],
+				[`is terribly frightened by ${player.pronouns[2]} own ghost and takes`, 1, '👻']
+			],
+			[
+				['is caught in a sudden windstorm, hit by debris and branches, and takes', 20, '🌬'],
+				['stumbles through some poison ivy and takes', 50, '🍂'],
```
RodEsp: How is stumbling through poison ivy or walking through saw grass 50 damage while getting struck by a meteor is only 1?

exploration/index.js
```diff
@@ -0,0 +1,191 @@
...
+		});
+	}
+
+	sendMonsterHome (explorer) {
+		if (this.monsterIsExploring(explorer.monster)) {
+			const explorerIndex = this.explorers.indexOf(explorer);
+
+			this.explorers.splice(explorerIndex, 1);
+
+			const deadMessage = `${explorer.monster.givenName}'s carcase is wheeled home in a cart by a kindly stranger`;
```
RodEsp: carcass?

exploration/discoveries/hazard.js
tweettypography: hahaha I think that's probability that you're seeing, not damage

exploration/discoveries/death.js
```diff
@@ -0,0 +1,34 @@
...
+	} = {}) {
+		super({ icon });
+	}
+
+	get stats () {
+		return this.flavor;
+	}
+
+	effect (player) { // eslint-disable-line class-methods-use-this
+		player.dead = true;
```
tweettypography: We usually do `.die(assailant);` rather than `.dead = true`. In this case assailant is "the world"

exploration/discoveries/hazard.js
```diff
@@ -0,0 +1,99 @@
...
+				[`is burnt by ${player.pronouns[2]} camp fire for`, 50, '🔥'],
+				['is caught in a sudden forest fire and takes', 20, '🔥'],
+				['steps on a hornet\'s nest and is stung repeatedly for', 40, '🐝'],
+				[`is caught off guard by a wild boar who, before dissapearing back into the foliage, rams ${player.pronouns[1]} for`, 70, '🐗'],
+				['eats a bad mushroom and takes', 30, '🍄'],
+				['is struck by a meteorite for', 1, '☄️']
+			]
+		];
+
+		return { hazards: damageFlavors[number] };
```
tweettypography: Wait, so what happens if you have damageDice higher than 4? There's no fallback right now so I could see this being a tricky bug in the future...

Also, to piggy-back on @RodEsp's question - Getting struck by a meteorite is only 4 damage??

exploration/discoveries/nothing.js
```diff
@@ -0,0 +1,52 @@
...
+	// Set defaults for these values that can be overridden by the options passed in
+	constructor ({
+		icon = '🤷‍'
+	} = {}) {
+		super({ icon });
+	}
+
+	get stats () {
+		return this.flavor;
+	}
```
tweettypography: How does this work? Can you write a test for it as an example?

exploration/index.js
```diff
@@ -0,0 +1,191 @@
...
+
+
+	set discoveries (discoveries) {
+		this.setOptions({
+			discoveries
+		});
+	}
+
+	getExplorer (targetMonster) {
+		return this.explorers.find(explorer => JSON.stringify(explorer.monster) === JSON.stringify(targetMonster));
```
tweettypography: Monster names have to be unique so you don't actually need such a deep and expensive comparison. Also, unless you're planning to save the state of explorations and are worried about mismatches after hydration -- the actual instances should be equal. That is, `explorer.monster` and `targetMonster` should actually be the exact same instance such that `explorer.monster === targetMonster` unless you've done a lot of work around that.

exploration/index.js
```diff
@@ -0,0 +1,191 @@
...
+				explorer
+			});
+
+			this.channelManager.queueMessage({
+				announce: `${monster.givenName} has gone exploring.
+
+The Road goes ever on and on
+Down from the town where it began.
+Now far ahead the Road has gone,
+And ${monster.givenName} must follow, if ${monster.pronouns[0]} can,
```
tweettypography: You forget, you've changed how pronouns work since you wrote this code :)

exploration/index.js
```diff
@@ -0,0 +1,191 @@
...
+				});
+		}, ONE_MINUTE);
+	}
+
+	makeDiscovery (explorer) {
+		const discoveries = shuffle(this.discoveries);
+
+		// if (explorer) {
+		// 	discoveries = discoveries.filter(discovery => explorer.canFind(discovery));
+		// }
```
tweettypography: So all discoveries are possible for everyone at this time?

exploration/index.js
```diff
@@ -0,0 +1,191 @@
...
+
+		const Discovery = discoveries.find(isProbable);
+
+		if (!Discovery) return this.makeDiscovery(explorer);
+
+		const discovery = new Discovery();
+		discovery.look(explorer.channel);
+		discovery.play(this.environment, explorer.monster);
```
tweettypography: Are these methods really synchronous?

game.js
```diff
@@ -437,6 +444,10 @@ class Game extends BaseClass {
 		return this.ring;
 	}
 
+	getExploration () {
+		return this.exploration;
+	}
+
 	lookAtRing (channel, ringName = 'main', showCharacters) {
 		return this.getRing(ringName).look(channel, showCharacters);
```
tweettypography: You should consider updating `getRing` to look for exploration as well as the "main" ring. Then all of the same ring methods could theoretically apply.

items/base.js
```diff
@@ -23,6 +23,12 @@ class BaseItem extends BaseClass {
 		return this.options.icon;
 	}
 
+	set icon (icon) {
+		this.setOptions({
+			icon
+		});
+	}
```
tweettypography: When do you do this? Also... I wish we could move the icons to the constructor instead of the options, I just haven't gotten around to it.

exploration/discoveries/hazard.js
cormacmccarthy: Correct, that's probability :D

exploration/index.js
cormacmccarthy: It's the British spelling, which I've been preferring when writing narrative because it adds a bit of a "High Fantasy" or "Classical Lit" flavor to the game ("catalogue" vs "catalog" for instance).

But... If it's just really throwing people off, I could be convinced to change it :)

exploration/discoveries/hazard.js
cormacmccarthy: https://news.nationalgeographic.com/news/2013/02/130220-russia-meteorite-ann-hodges-science-space-hit/

Yeah, I mean, it's just a little one... Or maybe it bounced first... alright, I'll update to clarify :P

game.js
cormacmccarthy: I'd have to look more deeply into this. It's not obvious to me that I'd want exploration to function as a ring...

items/base.js
cormacmccarthy: On line 23 in hazard.js I set `this.icon = this.flavor.icon`

If moving the icon to the constructor would still allow me to do that, that's fine with me.

exploration/index.js
cormacmccarthy: yes.

exploration/index.js
cormacmccarthy: I'll just reuse `canHold`...

exploration/index.js
cormacmccarthy: 🎉

exploration/index.js
cormacmccarthy: Oh sweet.

exploration/discoveries/nothing.js
cormacmccarthy: It _doesn't_ 🤦‍♂️

exploration/index.js
tweettypography: I guess I'm partly wondering why you have look followed by play - for both cards and items we usually show the card or item as part of play/use. The pattern in items, while more complicated than you need for this, is a good starting place. https://github.com/deck-monsters/deck-monsters/blob/master/items/base.js#L60

items/base.js
tweettypography: Interesting. This is fine for now

items/base.js
```diff
@@ -23,6 +23,12 @@ class BaseItem extends BaseClass {
 		return this.options.icon;
```
tweettypography: A good start would be `return this.options.icon || this.constructor.icon;` - that way all of the default icons could be on the constructor and we could reserve options for when you need to override.

exploration/discoveries/base.js
```diff
@@ -0,0 +1,29 @@
+const Promise = require('bluebird');
```
tweettypography: You don't seem to actually be using this

PR #157, opened 2018-01-09
Fix horn gore

tweettypography:


cards/pick-pocket.mocha.node.js
```diff
@@ -80,6 +81,33 @@ describe('./cards/pick-pocket.js', () => {
 			});
 	});
 
+	it('can run many times without error', () => {
+		const pickPocket = new PickPocketCard();
+		const player = randomContestant();
+
+		const promises = [];
+
+		for (let i = 0; i < 50; i++) {
```
cormacmccarthy: I'd kinda rather see each card tested rather than just hoping brute force will catch them all... But this will work for now if you don't feel like making that happen. You should just be able to pull in `cards/index.js:all` and use the for loop indices to fill the randomContestant's decks up with only the card at that indices.

tweettypography: Yeah this was just a quick and dirty way to catch the error you mentioned. Had a nice side effect of catching other errors I didn't expect to find either so I left it in. I do like your idea though, we should make a little testing helper for that which we can use with other scenarios where it might make sense to test multiple different card combinations.

PR #161, opened 2018-01-10
Adds survival knife card

iowillhoit:
2d4 hit, 2d4 heal if below 1/4 health


cards/survival-knife.js
```diff
@@ -0,0 +1,39 @@
...
+	} = {}) {
+		super({ icon, ...rest });
+
+		this.healCard = new HealCard({ healthDice: '2d4' });
+	}
+
+
+	get stats () {
+		return `${super.stats}
+Heal 2d4 if below a quarter health`;
```
tweettypography: Should be `Heal 2d4 instead...` right? Unless you want to change the logic so that it heals _and_ hits. In which case you'd definitely need to lower the probability as this would be quite powerful.

cards/survival-knife.js
```diff
@@ -0,0 +1,39 @@
...
+		...rest
+	} = {}) {
+		super({ icon, ...rest });
+
+		this.healCard = new HealCard({ healthDice: '2d4' });
+	}
+
+
+	get stats () {
+		return `${super.stats}
```
cormacmccarthy: It was unclear to me at first that this was an either/or. I thought it was "hit for 2d4 & heal for 2d4 if below a quarter health"

Perhaps:
```
return `Either ${super.stats} or
```

or something like that would make this clearer.

iowillhoit:
@tweettypography It's a knife that the pommel unscrews and it has matches, bandaids, fishing line, etc... inside of it.
survival knife

cormacmccarthy:
I immediately thought of one of those buck-knives that has a compass on the back and that you can store stuff inside of. If that's what you're referencing maybe we could dress it up a bit.

"If times get too rough, press the button on the hilt and stab yourself in the thigh for a Stimpak injection"

http://fallout.wikia.com/wiki/Stimpak_(Fallout_3)

Maybe this should be limited to FIGHTER class?

iowillhoit:
Yea, that's what I was referencing :)

cards/survival-knife.js
cormacmccarthy: Sorry, I know I suggested the Either or thing, but something like the following would actually maybe look better:
```
return `${super.stats}
- or -
Heal 2d4 if below a quarter health`;
```

cards/survival-knife.js
```diff
@@ -0,0 +1,39 @@
...
+		icon = '🗡',
+		...rest
+	} = {}) {
+		super({ icon, ...rest });
+
+		this.healCard = new HealCard({ healthDice: '2d4' });
+	}
+
+	get stats () {
+		return `Either ${super.stats}
```
cormacmccarthy: I'm not convinced you need this either here, especially since "Hit" will be capitalized.

iowillhoit: oh, missed that

cards/survival-knife.js
```diff
@@ -0,0 +1,39 @@
...
+
+class SurvivalKnifeCard extends HitCard {
+	// Set defaults for these values that can be overridden by the options passed in
+	constructor ({
+		icon = '🗡',
+		...rest
+	} = {}) {
+		super({ icon, ...rest });
+
+		this.healCard = new HealCard({ healthDice: '2d4' });
```
cormacmccarthy: Make this an option that you pass into the card in case someone wants to extend it later. See https://github.com/deck-monsters/deck-monsters/blob/master/cards/entrance.js#L26 for an example.

cards/survival-knife.mocha.node.js
```diff
@@ -0,0 +1,84 @@
...
+
+	after(() => {
+		pause.setTimeout.restore();
+	});
+
+	it('can be instantiated with defaults', () => {
+		const survivalKnife = new SurvivalKnifeCard();
+
+		expect(survivalKnife).to.be.an.instanceof(SurvivalKnifeCard);
+		expect(survivalKnife.stats).to.equal('Hit: 1d20 vs AC / Damage: 2d4\nHeal 2d4 if below a quarter health');
```
cormacmccarthy: This needs updated

cards/survival-knife.js
```diff
@@ -0,0 +1,42 @@
...
+
+class SurvivalKnifeCard extends HitCard {
+	// Set defaults for these values that can be overridden by the options passed in
+	constructor ({
+		icon = '🗡',
+		...rest
+	} = {}) {
+		super({ icon, ...rest });
+
+		this.healCard = new HealCard({ healthDice: this.healthDice });
```
tweettypography: Shouldn't this be `this.constructor.healDice`?

cormacmccarthy: oh, probably

cormacmccarthy: actually I think it should be `this.damageDice`

PR #162, opened 2018-01-10
Move card selection out of card equip

tweettypography:
Creates a separate method (used by equip) for selecting a card. This will also be used by the store when its ready. The new function supports selection of multiple cards at once.


items/base.js
```diff
@@ -43,33 +43,6 @@ class BaseItem extends BaseClass {
...
-		// a counter instead of boolean just in case the item can be used more than once
-		this.used += 1;
-
-		if (this.effect) {
-			return this.effect(character, monster);
-		}
-
-		return this.used;
-	}
-
```
cormacmccarthy: It's unclear to me why you moved this out of BaseItem into BasePotion. Seems like it could have been useful for all items...

tweettypography: Hm, good point. I'll put it back

cormacmccarthy:
This is going to save me quite a bit of work in exploration! 🎉

PR #165, opened 2018-01-11
New attributes & battle messages

cormacmccarthy:
Light rebalancing of monsters with INT in mind.

An all new Blast II card that utilizes INT instead of just level.

Updates across the board to take advantage of the new INT attr.An all new Blast II card that utilizes INT instead of just level

Updates across the board to take advantage of the new INT attr.

Tweaks Rolling & Rolled to make them much more scannable.


announcements/rolled.js
```diff
@@ -5,11 +5,17 @@ const announceRolled = (publicChannel, channelManager, className, monster, {
...
 	outcome
 }) => {
+	let rollResult = (roll.naturalRoll.result === 20) ? 'Natural 20!' : roll.result;
+	rollResult = (roll.naturalRoll.result === 1) ? 'Critical Failure!' : rollResult;
+	const vsMsg = vs ? ` vs ${vs}${target.icon}` : '';
+
 	publicChannel({
 		announce:
-`🎲  ${player.identity} rolled a ${roll.result} (natural ${roll.naturalRoll.result}${signedNumber(roll.bonusResult)}${signedNumber(roll.modifier)}) ${reason}
+`🎲 ${player.icon}${rollResult}${vsMsg}  ${player.identity} rolled a ${roll.result} (natural ${roll.naturalRoll.result}${signedNumber(roll.bonusResult)}${signedNumber(roll.modifier)}) ${reason}
```
tweettypography: Why isn't there a space between the player icon and the result? Same might go for between vs and the target icon but I'm not sure - maybe vs includes a space?

creatures/base.js
```diff
@@ -12,15 +12,28 @@ const pause = require('../helpers/pause');
...
 const HP_VARIANCE = 5;
-const MAX_AC_BOOST = (BASE_AC * 2) + AC_VARIANCE;
-const MAX_AC_MODIFICATION = 4;
-const MAX_ATTACK_BOOST = 10;
-const MAX_DAMAGE_BOOST = 6;
-const MAX_HP_BOOST = (BASE_HP * 2) + HP_VARIANCE;
-const MAX_HP_MODIFICATION = 12;
+// Do not access these directly. Get from this.getMaxModifications
+const MAX_PROP_MODIFICATIONS = {
+	ac: 1,
```
tweettypography: So the max AC modification used to be 4 and now it's 1 - what impact does that have?

tweettypography:
Let me know when you get a chance to fix conflicts with this.

creatures/base.js
cormacmccarthy: It's actually 1 + level now. So, for lower levels who have hardly any ac to begin with, you won't be able to take them down to almost nothing. Higher levels with lots of ac you'll be able to affect their ac a lot more.

Biggest problem I see with this is that at lower levels you also won't be able to _increase_ your ac through cards much. I'm considering making the increase cap be something like 1.5 or 2x the decrease floor. But that'll be a different PR. Possibly the one where I finally tweak boost (like I tweaked curse).

cormacmccarthy:
@tweettypography merged master.

announcements/rolled.js
cormacmccarthy: I was trying to keep it condensed, but after toying with it for a while this morning I've settled on a space. It now looks like this:

image


Here's all the stuff I tried (if you have some suggestions, feel free to play with it more):
image


Here's the final format I settled on:
```
:wave::game_die: :star2: *1d20 +4* _vs_ *12* :snake:  (_:star2: Dbb rolls 1d20 +4 vs :snake: King's AC (12) to determine if the hit was a success_)
```

PR #166, opened 2018-01-11
Add a store where you can sell cards and items

tweettypography:
Next tasks will be adding prices to all existing cards, writing some tests, and creating the 'buy' portion of the store.

### DIFF
--- a/items/store/sell.js
+++ b/items/store/sell.js
@@ -1,5 +1,7 @@
 const fantasyNames = require('fantasy-names');
+const random = require('lodash.random');
 const sample = require('lodash.sample');
+const throttle = require('lodash.throttle');
 
 const chooseCards = require('../../cards/helpers/choose');
 const chooseItems = require('../helpers/choose');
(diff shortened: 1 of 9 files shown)


items/store/sell.js
```diff
@@ -0,0 +1,95 @@
...
+
+				return chooseItems({ items, channel });
+			}
+
+			if (cards.length < 1) return Promise.reject(channel({ announce: "You don't have any cards." }));
+
+			return chooseCards({ cards, channel });
+		})
+		.then((choices) => {
+			const value = choices.reduce((total, choice) => total + Math.floor(choice.cost * 0.8), 0);
```
cormacmccarthy: I'd love to see this as somehow being a weighted semi-random number between 0.7 and 1.1 (with the likelihood of getting 0.7~0.9 being much more likely)

tweettypography: I thought about that but I think that people would just try to go back to the store over and over until they got a price they liked.

tweettypography: Maybe I could make the price change every few hours so if you wanted a better price you'd have to wait (but it could get worse)

cormacmccarthy: That seemed a lot harder, but would be ideal probably.

cormacmccarthy: I thought about people going back over and over, but I assumed that doing it more than a couple times just wouldn't be worth the effort.

cormacmccarthy:
You should probably also go through and update all the card costs while you're at it. They more or less all cost the same right now. Tiny tiny variations in prices...

items/store/sell.js
tweettypography: Done. You now get a new shop every two hours.

tweettypography:
@cormacmccarthy I'm going to add prices for all the cards before we start using this but I figured I'd do that in its own PR. We can collaborate on it too.

PR #167, opened 2018-01-12
Add prices to all cards

tweettypography:
Not all of these cards will actually be available for sale, of course, but this is the base price that will be used when you sell back to the store.

### DIFF
--- a/cards/enchanted-faceswap.js
+++ b/cards/enchanted-faceswap.js
@@ -60,10 +60,10 @@ class EnchantedFaceswapCard extends BaseCard {
 }
 
 EnchantedFaceswapCard.cardType = 'Enchanted Faceswap';
-EnchantedFaceswapCard.probability = 50;
-EnchantedFaceswapCard.description = 'A snapchat filter for the magically inclined. This spell will cause the next card played with the caster as the target to be reversed so that the player of the card becomes the target.';
-EnchantedFaceswapCard.cost = 4;
-EnchantedFaceswapCard.level = 1;
 EnchantedFaceswapCard.permittedClassesAndTypes = [CLERIC];
+EnchantedFaceswapCard.probability = 40;
+EnchantedFaceswapCard.description = 'A snapchat filter for the magically inclined. This spell will cause the next card played with the caster as the target to be reversed so that the player of the card becomes the target.';
+EnchantedFaceswapCard.level = 1;
+EnchantedFaceswapCard.cost = 25;
 
 module.exports = EnchantedFaceswapCard;
--- a/cards/fists-of-virtue.js
+++ b/cards/fists-of-virtue.js
@@ -24,6 +24,8 @@ FistsOfVirtueCard.cardType = 'Fists of Virtue';
 FistsOfVirtueCard.probability = 30;
 FistsOfVirtueCard.description = 'You strike at the biggest bully in the room.';
 FistsOfVirtueCard.level = 1;
+FistsOfVirtueCard.cost = 20;
+
 FistsOfVirtueCard.defaults = {
 	...HitCard.defaults,
 	damageDice: '1d8' // Slightly more possible damage
--- a/cards/kalevala.js
+++ b/cards/kalevala.js
@@ -56,9 +56,10 @@ It will now do ${this.damageDice} damage.`
 KalevalaCard.cardType = 'The Kalevala';
 KalevalaCard.probability = 10;
 KalevalaCard.description = 'Steadfast old Väinämöinen himself fashioned this instrument of eternal joy. Tune its pikebone pegs and it may lead you on to victory.'; // eslint-disable-line max-len
-KalevalaCard.cost = 6;
 KalevalaCard.level = 1;
+KalevalaCard.cost = 40;
 KalevalaCard.noBosses = true;
+
 KalevalaCard.defaults = {
 	...HitCard.defaults,
 	damageDice: damageLevels[0] // What begins weak may one day be strong
--- a/cards/vengeful-rampage.js
+++ b/cards/vengeful-rampage.js
@@ -15,10 +15,11 @@ Damage: ${this.damageDice} +1 per wound suffered`;
 }
 
 VenegefulRampageCard.cardType = 'Vengeful Rampage';
+VenegefulRampageCard.permittedClassesAndTypes = [BARBARIAN];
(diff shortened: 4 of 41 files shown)


cards/kalevala.js
```diff
@@ -56,9 +56,10 @@ It will now do ${this.damageDice} damage.`
 KalevalaCard.cardType = 'The Kalevala';
 KalevalaCard.probability = 10;
 KalevalaCard.description = 'Steadfast old Väinämöinen himself fashioned this instrument of eternal joy. Tune its pikebone pegs and it may lead you on to victory.'; // eslint-disable-line max-len
-KalevalaCard.cost = 6;
 KalevalaCard.level = 1;
+KalevalaCard.cost = 40;
```
iowillhoit: I feel like this should be more

cards/vengeful-rampage.js
```diff
@@ -15,10 +15,11 @@ Damage: ${this.damageDice} +1 per wound suffered`;
 }
 
 VenegefulRampageCard.cardType = 'Vengeful Rampage';
+VenegefulRampageCard.permittedClassesAndTypes = [BARBARIAN];
 VenegefulRampageCard.probability = 20;
 VenegefulRampageCard.description = 'Your wounds only make you stronger.';
 VenegefulRampageCard.level = 3;
-VenegefulRampageCard.permittedClassesAndTypes = [BARBARIAN];
+VenegefulRampageCard.cost = 40;
```
iowillhoit: Same here? This and the kalevala are really powerful

items/store/shop.js
```diff
@@ -0,0 +1,31 @@
...
+	'bright red',
+	'small, round'
+];
+
+let currentShop;
+const getShop = throttle(() => {
+	currentShop = {
+		name: fantasyNames('places', 'magic_shops'),
+		adjective: sample(ADJECTIVES),
+		priceOffset: random(0.6, 0.9),
```
iowillhoit: NICE

cards/kalevala.js
tweettypography: You won't be allowed to buy this card normally - this is what you could sell it for. I'm also planning that cards that you sell will go into the store (which is reset every two hours) so if you _did_ sell your Kalevala someone else would have a window of time in which they could buy it, but I'm not sure that anyone would really want to do that.

cards/vengeful-rampage.js
tweettypography: At least for now you probably won't be able to buy these either.

cormacmccarthy:
I'd love to see some sort of formula used to calculate price. Something along the lines of:

(level * rarity * maxDamage || maxHeal || maxCurse) + priceModifier

priceModifier would be something you'd set in the options to jack up (or down) the price when the formula just isn't quite cutting it.

tweettypography:
I guess we could come up with something like that but how would it price Faceswap? The Kalevala? The "epic" cards that are all level 0? I think for now defining them by hand is best

cards/enchanted-faceswap.js
```diff
@@ -60,10 +60,10 @@ class EnchantedFaceswapCard extends BaseCard {
...
 
 EnchantedFaceswapCard.cardType = 'Enchanted Faceswap';
-EnchantedFaceswapCard.probability = 50;
+EnchantedFaceswapCard.permittedClassesAndTypes = [CLERIC];
+EnchantedFaceswapCard.probability = 40;
 EnchantedFaceswapCard.description = 'A snapchat filter for the magically inclined. This spell will cause the next card played with the caster as the target to be reversed so that the player of the card becomes the target.';
-EnchantedFaceswapCard.cost = 4;
 EnchantedFaceswapCard.level = 1;
-EnchantedFaceswapCard.permittedClassesAndTypes = [CLERIC];
+EnchantedFaceswapCard.cost = 25;
```
cormacmccarthy: Doesn't feel like enough...

Cloak of invisibility just hides you, whereas this not only hides you, but reflects your enemies' attack back at them. Feels like it should be at least 50, if not 60.

cards/fists-of-virtue.js
```diff
@@ -24,6 +24,8 @@ FistsOfVirtueCard.cardType = 'Fists of Virtue';
 FistsOfVirtueCard.probability = 30;
 FistsOfVirtueCard.description = 'You strike at the biggest bully in the room.';
 FistsOfVirtueCard.level = 1;
+FistsOfVirtueCard.cost = 20;
```
cormacmccarthy: Can we make this at least 25? It's actually a decent card for low level 1:1 battles because it hits for 1d8.

cormacmccarthy:
The problem is that it is really hard to keep all the prices in my head to compare them against each other to make sure its reasonable.

When I was trying to evaluate I realized I wanted to make a spreadsheet that tracked:

- card level
- probability
- max Possible Damage
- max Possible Heal
- max Avg Damage
- max Avg Heal
- max turns hidden
- avg turns hidden
- max turns immobilized
- avg turns immobilized
- min turns immobilized
- number of players targeted

And then give each of those a "weight" and a multiplier and add it all up to come up with a base-line cost for each card that we could add/subtract to/from with one final override number that lets us tweak it a little bit if necessary. Which basically just describes a formula that maybe the shop can use to price cards.

Faceswap does a max possible damage of the hardest hitting card (BattleFocus, which can hit for something like 25 if all the stars align because it caps out at maxHP/2 I think). It does an average damage of the average of the average damage of all cards modified by their probability.

This is no simple formula for sure... And maybe it would actually just be better done as a master spreadsheet that you plug your card into and plug in all the values in order to get your cost suggestion.

cards/fists-of-virtue.js
tweettypography: Makes sense, I’ll increase it in the next PR

tweettypography:
That should be possible so we could look into it for the future (though you can make a spreadsheet now if you’d like). If we added those values to all cards (hard coded in some cases, calculated in others) then we could probably come up with a reasonable base formula. The thing is, that might make reading and editing these values even harder so maybe a spreadsheet is still better?

cards/enchanted-faceswap.js
tweettypography: Fair enough - just kept it lower so you couldn’t sell it for a lot but since it is for sale we probably need to increase it.

PR #191, opened 2018-01-16
Spawn gives a bit more guidance when spawning monsters

cormacmccarthy:
Gives more specific signals around character spawning, including suggesting names and "colors", to help avoid the confusion that currently happens when someone is being asked the spawn questions.

Makes gender the second question you are asked since you are asked to make historically gender-based decisions (name, color) _before_ being asked gender currently.

The changes in weeping angel protect against "a" being followed by a word that begins with a vowel, eg: "a auburn".

The changes in battlefield cause an actual boss to be spawned instead of the left-over pre-boss "boss" we used to spawn.

### DIFF
--- a/monsters/helpers/spawn.js
+++ b/monsters/helpers/spawn.js
@@ -59,8 +59,8 @@ ${getCreatureTypeChoices(allMonsters)}`,
 			let question = '';
 			if (alreadyTaken) question += 'That name is already taken, please choose a different name. ';
 
-			const name1 = names(Monster.creatureType.toLowerCase(), options.gender, monsterNames);
-			const name2 = names(Monster.creatureType.toLowerCase(), options.gender, [ name1, ...monsterNames ]);
+			const name1 = names(Monster.creatureType, options.gender, monsterNames);
+			const name2 = names(Monster.creatureType, options.gender, [ name1, ...monsterNames ]);
 
 			question += `What would you like to name ${PRONOUNS[options.gender].him}? ${name1}? ${name2}? Something else?`;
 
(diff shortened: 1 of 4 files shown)


monsters/helpers/spawn.js
```diff
@@ -77,8 +84,32 @@ ${getCreatureTypeChoices(allMonsters)}`,
...
+				case CREATURE_TYPES.MINOTAUR.toLowerCase():
+					example = 'scarred, wrinkled, and beautifully auburn';
+					descriptor = 'skin and hair look like';
+					break;
+				case CREATURE_TYPES.GLADIATOR.toLowerCase():
+					example = 'tattered rags';
+					descriptor = 'garments look like';
+					break;
+				case CREATURE_TYPES.WEEPING_ANGEL.toLowerCase():
+					example = 'deceptively glorious';
```
tweettypography: This doesn't make sense now that you added a "color flavor" - for example you could end up getting a "gloriously deceptively glorious" angel. I'm not sure I really like the color flavors for this reason.

monsters/helpers/spawn.js
```diff
@@ -55,7 +59,10 @@ ${getCreatureTypeChoices(allMonsters)}`,
 			let question = '';
 			if (alreadyTaken) question += 'That name is already taken, please choose a different name. ';
 
-			question += `What would you like to name your new ${Monster.creatureType.toLowerCase()}?`;
+			const name1 = names(Monster.creatureType.toLowerCase(), options.gender, monsterNames);
```
tweettypography: Why are you using `toLowerCase` here?

monsters/helpers/spawn.js
cormacmccarthy: Yeah I debated between this and some sort of a/an solution. I’ll do the a/an...

PR #192, opened 2018-01-16
Adds a new "Jinn" monster type

tweettypography:
Adds a new class "Bard" with a new monster of that class "Jinn" a new Epic card for that monster "Sandstorm" and a new class-specific card "Bad Batch"

### DIFF
--- a/cards/bad-batch.js
+++ b/cards/bad-batch.js
@@ -45,7 +45,7 @@ class BadBatchCard extends BaseCard {
 						const healRoll = getHealRoll.call(card, player);
 
 						this.emit('rolled', {
-							reason: 'to determine how much to heal',
+							reason: 'to measure out a shot of whiskey.',
 							card: this,
 							roll: healRoll,
 							player,
@@ -88,7 +88,7 @@ BadBatchCard.flavors = {
 		['slips a mysterious liquid in the cup of', 60],
 		['buys a round for', 60],
 		['offers a drink to', 60],
-		['accidentally gives lead poisoning to', 20],
+		['"accidentally" gives lead poisoning to', 20],
 		['pours acid straight down the throat of', 5]
 	]
 };
--- a/cards/bad-batch.js
+++ b/cards/bad-batch.js
@@ -1,0 +1,96 @@
+/* eslint-disable max-len */
+const BaseCard = require('./base');
+const WhiskeyShotCard = require('./whiskey-shot');
+const ScotchCard = require('./scotch');
+
+const { BARD } = require('../helpers/classes');
+
+const EFFECT_TYPE = 'BadBatchEffect';
+
+class BadBatchCard extends BaseCard {
+	// Set defaults for these values that can be overridden by the options passed in
+	constructor ({
+		icon = '🍻'
+	} = {}) {
+		super({ icon });
+	}
+
+	get targetCards () {
+		return this.constructor.targetCards;
+	}
+
+	get stats () {
+		return `The next ${this.targetCards.join(' or ')} played will poison rather than heal.`;
+	}
+
+	getTargets (player) { // eslint-disable-line class-methods-use-this
+		return [player];
+	}
+
+	effect (badBatchPlayer, badBatchTarget, badBatchRing) {
+		const badBatchEffect = ({
+			card
+		}) => {
+			if (this.targetCards.includes(card.cardType)) {
+				badBatchRing.encounterEffects = badBatchRing.encounterEffects.filter(encounterEffect => encounterEffect !== badBatchEffect);
+
+				const { effect, getHealRoll } = card;
+
+				if (effect && getHealRoll) {
+					card.effect = (player, target) => {
+						this.emit('narration', {
+							narration: `${target.givenName} doesn't notice that the seal on ${target.pronouns.his} bottle has been tampered with.`
+						});
+
+						const healRoll = getHealRoll.call(card, player);
+
+						this.emit('rolled', {
+							reason: 'to determine how much to heal',
+							card: this,
+							roll: healRoll,
+							player,
+							target,
+							outcome: 'Poisoned!'
+						});
+
+						return target.hit(healRoll.result, badBatchTarget, this);
+					};
+				}
+			}
+
+			return card;
+		};
+
+		badBatchEffect.effectType = EFFECT_TYPE;
+
+		badBatchRing.encounterEffects = [...badBatchRing.encounterEffects, badBatchEffect];
+
+		this.emit('narration', {
+			narration: `${badBatchTarget.identity} brews up a ${this.icon} bad batch of the strong drink.`
+		});
+
+		return true;
+	}
+}
+
+BadBatchCard.cardType = 'Bad Batch';
+BadBatchCard.permittedClassesAndTypes = [BARD];
+BadBatchCard.targetCards = [WhiskeyShotCard.cardType, ScotchCard.cardType];
+BadBatchCard.probability = 30;
+BadBatchCard.description = 'Nothing like a little bathtub moonshine stored in sturdy lead jugs.';
+BadBatchCard.level = 1;
+BadBatchCard.cost = 30;
+BadBatchCard.notForSale = true;
+
+BadBatchCard.flavors = {
+	hits: [
+		['poisons', 80],
+		['slips a mysterious liquid in the cup of', 60],
+		['buys a round for', 60],
+		['offers a drink to', 60],
+		['accidentally gives lead poisoning to', 20],
+		['pours acid straight down the throat of', 5]
+	]
+};
+
+module.exports = BadBatchCard;
--- a/cards/enthrall.mocha.node.js
+++ b/cards/enthrall.mocha.node.js
@@ -10,7 +10,7 @@ const pause = require('../helpers/pause');
 
 
 const {
-	GLADIATOR, MINOTAUR, BASILISK, WEEPING_ANGEL
+	GLADIATOR, JINN, MINOTAUR, BASILISK, WEEPING_ANGEL
 } = require('../helpers/creature-types');
 
 describe('./cards/enthrall.js', () => {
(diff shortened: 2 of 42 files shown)


cards/heal.js
```diff
@@ -51,22 +51,23 @@ Possiblity of Stroke of Luck`;
 			healResult = Math.floor(player.maxHp / 2);
 			this.emit('narration', {
 				narration: `Stoke of Luck!
-Wait... wasn't this the questionable phial you found on the floor behind the shelf? Is it safe? Desperate times... Down the hatch!`
+Wait... wasn't this the questionable vial you found on the floor behind the shelf? Is it safe? Desperate times... Down the hatch!`
```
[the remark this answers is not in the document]
tweettypography: No, I thought it was just a typo but now I realize you used that outdated spelling on purpose to go with the theme.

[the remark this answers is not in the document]
tweettypography: Feel free to push a change right back to the branch if you want to switch it back!

cards/bad-batch.js
```diff
@@ -0,0 +1,96 @@
...
+BadBatchCard.cost = 30;
+BadBatchCard.notForSale = true;
+
+BadBatchCard.flavors = {
+	hits: [
+		['poisons', 80],
+		['slips a mysterious liquid in the cup of', 60],
+		['buys a round for', 60],
+		['offers a drink to', 60],
+		['accidentally gives lead poisoning to', 20],
```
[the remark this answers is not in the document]
cormacmccarthy: ROFL, actually maybe just throw some scare-quotes around accidentally:

"accidentally" gives lead poisoning to

cards/enthrall.mocha.node.js
```diff
@@ -53,9 +54,9 @@ Chance to immobilize your opponents with your shocking beauty.`;
 		expect(enthrall.hitOnFail).to.be.false;
 		expect(enthrall.doDamageOnImmobilize).to.be.false;
 		expect(enthrall.stats).to.equal(stats);
-		expect(enthrall.strongAgainstCreatureTypes).to.deep.equal([GLADIATOR, BASILISK]);
+		expect(enthrall.strongAgainstCreatureTypes).to.deep.equal([BASILISK, GLADIATOR]);
 		expect(enthrall.weakAgainstCreatureTypes).to.deep.equal([MINOTAUR, WEEPING_ANGEL]);
-		expect(enthrall.permittedClassesAndTypes).to.deep.equal([WEEPING_ANGEL]);
+		expect(enthrall.permittedClassesAndTypes).to.deep.equal([JINN, WEEPING_ANGEL]);
```
cormacmccarthy: You added JINN to permitted, but didn't add a case for uselessAgainst

cards/entrance.mocha.node.js
```diff
@@ -53,9 +54,9 @@ Chance to immobilize and damage your opponents with your painfully shocking beau
 		expect(entrance.hitOnFail).to.be.true;
 		expect(entrance.doDamageOnImmobilize).to.be.true;
 		expect(entrance.stats).to.equal(stats);
-		expect(entrance.strongAgainstCreatureTypes).to.deep.equal([GLADIATOR, BASILISK]);
+		expect(entrance.strongAgainstCreatureTypes).to.deep.equal([BASILISK, GLADIATOR]);
 		expect(entrance.weakAgainstCreatureTypes).to.deep.equal([MINOTAUR, WEEPING_ANGEL]);
-		expect(entrance.permittedClassesAndTypes).to.deep.equal([WEEPING_ANGEL]);
+		expect(entrance.permittedClassesAndTypes).to.deep.equal([JINN, WEEPING_ANGEL]);
```
cormacmccarthy: same as enthrall, no uselessAgainst param

cards/sandstorm.js
```diff
@@ -0,0 +1,134 @@
...
+			return super.effect(sandstormPlayer, sandstormTarget) && super.effect(sandstormPlayer, sandstormTarget);
+		}
+
+		const sandstormEffect = ({
+			card,
+			phase,
+			player: effectPlayer
+		}) => {
+			if (phase === ATTACK_PHASE && effectPlayer === sandstormTarget) {
+				sandstormTarget.encounterEffects = sandstormTarget.encounterEffects.filter(encounterEffect => encounterEffect !== sandstormEffect);
```
cormacmccarthy: I'd like to see this be semi-permanent (like immobilize) with a saving throw:

Legolost makes saving throw 1d20 + spell bonus (12 +2) vs Robin Williams's int + spell bonus (12) to see if he can counteract the sandstorm the Jinn is kicking up.
🎲 14 v 12
Success! Legolost pulls his shirt up over his mouth, pulls his goggles out of his pocket, and _gets to work_.


If stroke of luck, he should escape entirely. Curse of loki, he just hits himself or gives away his heal or boost card. Otherwise, proceed as programmed.

Especially because this doesn't immobilize so much as just causes-retargets and general confusion.

cards/sandstorm.js
```diff
@@ -0,0 +1,134 @@
...
+			}
+
+			return card;
+		};
+
+		sandstormEffect.effectType = EFFECT_TYPE;
+
+		sandstormTarget.encounterEffects = [...sandstormTarget.encounterEffects, sandstormEffect];
+
+		return super.effect(sandstormPlayer, sandstormTarget);
```
cormacmccarthy: Because of the saving throw weakening the card a little, I think this should happen _each round_ the player is in the sand storm in addition to (maybe instead of?) the initial hit. At the least there should be room for some sort of ongoing damage in case we want to make Sandstorm II (which, I really like this card, and would like to see a higher level version that lasts longer and does ongoing damage).

monsters/jinn.js
```diff
@@ -0,0 +1,67 @@
...
+	'tip-toes',
+	'hides'
+];
+
+class Jinn extends BaseMonster {
+	constructor (options) {
+		const defaultOptions = {
+			dexModifier: 1,
+			strModifier: 0,
+			intModifier: 1,
```
cormacmccarthy: I tried to give everyone a net of 3. Jinn only has two, but once I implement speed that can be their 3rd (and that actually makes sense), so this is probably fine.

cards/bad-batch.js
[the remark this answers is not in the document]
tweettypography: Ah that's good - it was obviously tongue-in-cheek but the quotes make it better :)

cards/flee.js
```diff
@@ -41,7 +41,7 @@ class FleeCard extends BaseCard {
 					this.emit('stay', {
 						fleeResult: fleeRoll.result,
 						fleeRoll,
-						player,
+						player: target,
```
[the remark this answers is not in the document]
tweettypography: Maybe. Sometimes that can get switched by modifying cards. Generally speaking the "target" is always the one who is acted upon (even if that is the player)

cards/sandstorm.js
tweettypography: Do you think this might be overpowered though?

cards/sandstorm.js
tweettypography: I like the idea that the initial hit always happens - it's the damage from the force of the initial sandstorm whipping up. I like the idea of a level II that does ongoing damage (like coil) though.

monsters/jinn.js
tweettypography: Basilisk, Minotaur, and Weeping Angel all have 2, it's only Gladiators that have 3 right now. I might just change gladiators to match.

cards/bad-batch.js
```diff
@@ -0,0 +1,96 @@
...
+				if (effect && getHealRoll) {
+					card.effect = (player, target) => {
+						this.emit('narration', {
+							narration: `${target.givenName} doesn't notice that the seal on ${target.pronouns.his} bottle has been tampered with.`
+						});
+
+						const healRoll = getHealRoll.call(card, player);
+
+						this.emit('rolled', {
+							reason: 'to determine how much to heal',
```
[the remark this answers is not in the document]
tweettypography: Done

cards/enthrall.mocha.node.js
tweettypography: Done

cards/entrance.mocha.node.js
tweettypography: Done

cards/sandstorm.js
tweettypography: I'm not going to make any changes for this at the moment as it'll be easy to implement when we make a level II

monsters/jinn.js
tweettypography: Done

cards/sandstorm.js
tweettypography: I'm assuming you mean a saving throw each round, right? Not sure what the difference would be between counteracting the store and "escaping entirely" for the stroke of luck either. I'm leaning toward leaving it as-is for now and seeing how it plays out in the real world. If we need to change it we can, but otherwise we'll implement something like this in the level II

cormacmccarthy: When I said "escaping entirely" it was under the assumption that you were trapped in the storm for N rounds. A successful saving throw would just keep you safe that round. A nat 20 would end the storm early.

Whenever we enact a status effect we should give the target a chance to make a saving throw, either at the moment the card is played (Blink) or at the time that the card's effect goes into play (Immobilize). Otherwise the card ends up being OP, or at the very least feeling unfair (Invisibility & Faceswap, which I'm working on fixing). When there's no saving throw it feels like Deus Ex Machina just came and smacked you.

monsters/jinn.js
cormacmccarthy: haha, you're right. Trying to remember why I gave Gladiators 3 🤔 . It may have been a mix of Tradition for humans and a little boost because it seemed like they might not have many good cards specific to them.

cards/sandstorm.js
tweettypography: Makes sense. Right now it only effects the next play, it doesn't extend for any significant amount of time. Because it only mixes up targeting it's less overwhelmingly powerful than you might think - in tests it was fairly common that you'd just end up playing your card against the person you would have originally targeted anyway.

PR #193, opened 2018-01-17
Flesh out heal stats

cormacmccarthy:
Cards should display all of their mechanics. 

Imagine you are at a table and you play the card. It needs to tell you everything you need to know in order to actually play it.

I need to go through and audit all of the cards...


cards/heal.js
```diff
@@ -23,7 +23,10 @@ class HealCard extends BaseCard {
 
 	get stats () {
 		return `Health: ${this.healthDice}
-Possiblity of Stroke of Luck`;
++ spell bonus (diminished by 1 each use until 0, then resets)
+
+1% chance to heal half max hp
+1% chance to heal to poison`;
```
tweettypography: Should be "1% chance to poison"

PR #194, opened 2018-01-17
Revert "Add the sandstorm card to all players"

tweettypography:
This reverts commit 3ea880bebf48ffbf484d31827727d4f4a0f527a8 so that the sandstorm card isn't re-added on every restart (now that all players should have it).

### DIFF
--- a/characters/beastmaster.js
+++ b/characters/beastmaster.js
@@ -10,7 +10,7 @@ const { monsterCard } = require('../helpers/card');
 const { spawn, equip } = require('../monsters');
 const TENSE = require('../helpers/tense');
 
-const DEFAULT_MONSTER_SLOTS = 7;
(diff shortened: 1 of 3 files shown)


characters/beastmaster.js
```diff
@@ -10,7 +10,7 @@ const { monsterCard } = require('../helpers/card');
 const { spawn, equip } = require('../monsters');
 const TENSE = require('../helpers/tense');
 
-const DEFAULT_MONSTER_SLOTS = 7;
+const DEFAULT_MONSTER_SLOTS = 6;
```
cormacmccarthy: why?

tweettypography: Ah, that was just mixed in with the other changes. I'll fix

PR #201, opened 2018-01-19
Targeting Scrolls

tweettypography:


items/scrolls/cobra_kai.js
```diff
@@ -0,0 +1,25 @@
...
+
+class CobraKaiScroll extends TargetingScroll {
+	// Set defaults for these values that can be overridden by the options passed in
+	constructor ({
+		icon = '🐍'
+	} = {}) {
+		super({ icon });
+	}
+
+	getTargetingDetails (monster) { // eslint-disable-line class-methods-use-this
```
cormacmccarthy: we are possibly inconsistent around what we call "monster"... Near as I can tell, generally we refer to the acting creature as the "player" and the target creature as the "monster".

items/scrolls/cobra_kai.js
```diff
@@ -0,0 +1,25 @@
...
+		return `${monster.givenName} will target the player with the lowest current xp while ${monster.pronouns.he} is in the ring unless directed otherwise by a specific card.`;
```
cormacmccarthy: I'm assuming there will be some methodology for exposing these details to the user when they look at the scroll before they use it. Correct?

items/scrolls/cobra_kai.js
tweettypography: Both the `player` and the `target` are monsters. Unlike a card, this just acts on a single monster so we don't have to split them up that way

tweettypography: In a sense the `player` for items is actually the character

items/scrolls/cobra_kai.js
tweettypography: I just added something to the description for now `You target the weakest player in the ring, every time.` I agree that we should have a way to preview this though, just have to come up with commands that make sense.

PR #202, opened 2018-01-19
Rolls should message more appropriately when you hit yourself

tweettypography:
Improves readability of roll message a little bit, especially in cases where you accidentally target yourself. Another good improvement would be an _even worse_ critical fail when you are already confused and targeting yourself and you roll a 1.

### DIFF
--- a/cards/hit.js
+++ b/cards/hit.js
@@ -49,12 +49,18 @@ class HitCard extends BaseCard {
 			commentary = 'Miss... Tie goes to the defender.';
 		}
 
+		let reason;
+		if (player === target) {
+			reason = `vs ${target.pronouns.his} own ${this.targetProp.toLowerCase()} (${target[this.targetProp]}) in confusion.`;
+		} else {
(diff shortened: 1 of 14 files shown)


cards/hit.js
```diff
@@ -49,12 +49,18 @@ class HitCard extends BaseCard {
 			commentary = 'Miss... Tie goes to the defender.';
 		}
 
+		let reason;
+		if (player === target) {
+			reason = `vs ${target.pronouns.his} own ${this.targetProp.toLowerCase()} (${target[this.targetProp]}) in confusion.`;
+		} else {
+			reason = `vs ${target.givenName}'s ${this.targetProp.toLowerCase()} (${target[this.targetProp]}) to determine if the hit was a success.`;
+		}
+
```
cormacmccarthy: All these descendants of `hit` seem like they could benefit from a `getHitCheckReason` or something. Haven't really evaluated all of the code to see if it really _is_ repeated as much as I think it is though.

PR #203, opened 2018-01-19
Add a couple more scrolls

tweettypography:


items/scrolls/la-carambada.js
```diff
@@ -0,0 +1,25 @@
...
+class LaCarambadaScroll extends TargetingScroll {
+	// Set defaults for these values that can be overridden by the options passed in
+	constructor ({
+		icon = '🐍'
+	} = {}) {
+		super({ icon });
+	}
+
+	getTargetingDetails (monster) { // eslint-disable-line class-methods-use-this
+		return `${monster.givenName} will look for the living opponent with the highest possible hp while ${monster.pronouns.he} is in the ring and target them, even if that player currently has less hp, unless directed otherwise by a specific card.`;
```
cormacmccarthy: Wait... What? At the beginning of the battle they'll target the player with the highest HP (other than themselves) and then keep targeting that player until that player is dead? Or they'll keep re-targeting to the player with the highest HP (who is not themselves --just like the fists of virtue card)?

tweettypography: They'll target the player with the highest max hp (who is still alive). So yes, basically the player with the most hp at the beginning of the battle. I actually have the wrong strategy associated with this one right now though so it's good you asked, let me fix :)

The one that's like fists of virtue and which is currently on this card but which we will create another item for is `TARGET_HIGHEST_HP_PLAYER` whereas the one intended for this item is `TARGET_MAX_HP_PLAYER`

cormacmccarthy: I feel like something like this may be more easily parsed:

```
return `Unless directed otherwise by a specific card, ${monster.givenName} will target whichever living opponent would have the highest hp if they were at full health.`;
```

items/scrolls/sir-robin.js
```diff
@@ -0,0 +1,25 @@
...
+class SirRobinScroll extends TargetingScroll {
+	// Set defaults for these values that can be overridden by the options passed in
+	constructor ({
+		icon = '🐍'
+	} = {}) {
+		super({ icon });
+	}
+
+	getTargetingDetails (monster) { // eslint-disable-line class-methods-use-this
+		return `whenver ${monster.givenName} is in the ring ${monster.pronouns.he} will bravely look about, choose the player with the highest current hp, and target them, unless directed otherwise by a specific card.`;
```
cormacmccarthy: typo "whenver"

items/scrolls/sir-robin.js
```diff
@@ -0,0 +1,25 @@
+/* eslint-disable max-len */
+
+const TargetingScroll = require('./targeting');
+const { TARGET_HIGHEST_HP_PLAYER } = require('../../helpers/targeting-strategies');
+
+class SirRobinScroll extends TargetingScroll {
+	// Set defaults for these values that can be overridden by the options passed in
+	constructor ({
+		icon = '🐍'
```
cormacmccarthy: Why a snake?

items/scrolls/la-carambada.js
```diff
@@ -0,0 +1,25 @@
+/* eslint-disable max-len */
+
+const TargetingScroll = require('./targeting');
+const { TARGET_MAX_HP_PLAYER } = require('../../helpers/targeting-strategies');
+
+class LaCarambadaScroll extends TargetingScroll {
+	// Set defaults for these values that can be overridden by the options passed in
+	constructor ({
+		icon = '🐍'
```
cormacmccarthy: Why a snake?

items/scrolls/chaos-theory.js
```diff
@@ -0,0 +1,25 @@
+/* eslint-disable max-len */
+
+const TargetingScroll = require('./targeting');
+const { TARGET_RANDOM_PLAYER } = require('../../helpers/targeting-strategies');
+
+class ChaosTheoryScroll extends TargetingScroll {
+	// Set defaults for these values that can be overridden by the options passed in
+	constructor ({
+		icon = '🐍'
```
cormacmccarthy: how about 🦋

PR #204, opened 2018-01-19
Add The Annals of Qin Shi Huang scroll

tweettypography:


items/scrolls/qin-shi-huang.js
```diff
@@ -0,0 +1,25 @@
...
+
+	getTargetingDetails (monster) { // eslint-disable-line class-methods-use-this
+		return `${monster.givenName} will seek to consolidate ${monster.pronouns.his} power and lay waste to ${monster.pronouns.his} biggest foes in the ring by targeting the opponent with the highest xp, unless directed otherwise by a specific card.`;
+	}
+}
+
+QinShiHuangScroll.itemType = 'The Annals of Qin Shi Huang';
+QinShiHuangScroll.description = `焚書坑儒
+
+Target your opponent with the highest xp.`;
```
cormacmccarthy: hahaha, if you're going to use xp to target an opponent, you better use _the highest xp_ you can!

Maybe "the" instead of "your"?

items/scrolls/targeting.js
```diff
@@ -33,7 +33,7 @@ class TargetingScroll extends BaseScroll {
 			monster.targetingStrategy = targetingStrategy;
 		}
 
-		let narration = `${monster.givenName} learns new tactics from an ancient scroll entitled _${this.itemType}_.`;
+		let narration = `${monster.givenName} learns new tactics from an 📜 well-worn scroll entitled _${this.itemType}_.`;
```
cormacmccarthy: "an well-worn"

Also, this kind of makes me want to have it actually be `from ${this.agedness} scroll entitled` so we could put anything from "an ancient" to "an extremely new" for flavor depending on what fits with the feel of the scroll.

items/scrolls/qin-shi-huang.js
```diff
@@ -0,0 +1,25 @@
+/* eslint-disable max-len */
+
+const TargetingScroll = require('./targeting');
+const { TARGET_HIGHEST_XP_PLAYER } = require('../../helpers/targeting-strategies');
+
+class QinShiHuangScroll extends TargetingScroll {
+	// Set defaults for these values that can be overridden by the options passed in
+	constructor ({
+		icon = '🔥'
```
cormacmccarthy: Unless something about the history associated with this means "fire" it could be more fun to have Kanji here instead. `'焚'`

tweettypography: It's the "burning of books and burying of scholars" that I'm referencing with the emoji. I don't mind switching to the character though as it has basically the same meaning as the emoji and also strengthens the tie to China. I thought of using 🥢 but it seemed too benign.

items/scrolls/targeting.js
tweettypography: Sounds fun. I'm only changing this here because I used ancient twice before

PR #208, opened 2018-01-20
...according to clever hans

cormacmccarthy:
It's like the "with a hammer" of scrolls.

And actually, the name serves a double meaning, because if you're _actually very clever_ you _could_ use some of these scrolls to success.


helpers/targeting-strategies.js
```diff
@@ -78,6 +78,9 @@ function getTarget ({ playerContestant, contestants = [], strategy = TARGET_NEXT
 				return potentialTarget;
 			}, defaultTarget);
 		}
+		case TARGET_PLAYER_WHO_HIT_YOU_LAST_ACCORDING_TO_HANS: {
+			getTarget({ playerContestant, contestants, TARGET_PLAYER_WHO_HIT_YOU_LAST, false });
+		}
```
tweettypography: Should have a “return” and your variables need to be named, but otherwise this looks good.

cormacmccarthy:
A few unit tests and this is done (I want to actually add some unit tests for the clever hans targeting strategies).

helpers/targeting-strategies.mocha.node.js
```diff
@@ -4,7 +4,7 @@ const targetingStrategies = require('./targeting-strategies');
 
 const HitCard = require('../cards/hit');
 
-describe('./helpers/targeting-strategies.js', () => {
+describe.only('./helpers/targeting-strategies.js', () => {
```
tweettypography: .only

tweettypography:
You'll want to merge master and apply the new probability / cost logic to these. Also, I'm assuming we should mark them as "not for sale" so that they can be randomly given out as "rewards" but won't actually clog up the inventory in the store - they should be more unique and special than that.

cards/enthrall.mocha.node.js
```diff
@@ -80,7 +80,7 @@ Chance to immobilize your opponents with your shocking beauty.`;
 
 		expect(enthrall.getFreedomThreshold(player, target)).to.equal(10 + enthrall.freedomThresholdModifier);
 
-		target.encounterModifiers = { pinnedTurns: 2 };
+		target.encounterModifiers.pinnedTurns = 2;
```
tweettypography: I feel like we need to get this in ASAP

PR #211, opened 2018-01-23
Improve level text

tweettypography:
Also fix the display of heal results


cards/heal.js
```diff
@@ -46,45 +47,67 @@ class HealCard extends BaseCard {
 		return roll({ primaryDice: this.healthDice, modifier: player.encounterModifiers.healModifier, bonusDice: player.bonusIntDice });
 	}
 
-	checkLuck () { // eslint-disable-line class-methods-use-this
+	checkSuccess (healRoll, target) { // eslint-disable-line class-methods-use-this
+		const hundred = random(1, 100);
+
+		const strokeOfLuck = (hundred === 7);
+		const curseOfLoki = (hundred === 13);
```
cormacmccarthy: 😂

cards/heal.js
```diff
@@ -46,45 +47,67 @@ class HealCard extends BaseCard {
...
 		if (strokeOfLuck) {
-			healResult = Math.floor(target.maxHp / 2);
 			this.emit('narration', {
 				narration: `Stoke of Luck!
 Wait... wasn't this the questionable phial you found on the floor behind the shelf? Is it safe? Desperate times... Down the hatch!`
 			});
 		// Curse of Loki
-		} else if (curseOfLoki) {
-			healResult *= -1;
+		} else if (success && curseOfLoki) {
```
cormacmccarthy: 🤔 It's not obvious to me how success could be false, but curseOfLoki could be true.

This _looks_ like a bug, because one could easily assume that `curseOfLoki` would _not_ be a success condition.

cormacmccarthy: oh, now I understand. I feel like this is much more clear:

`} else if (result < 0) {`

tweettypography: That would work but it would actually be different behavior. `success` is a roll above 0, basically how much you're going to drink. If your bottle is empty you'll never know that it would have been poison. We could change the Curse of Loki text so that you'll _know_ it was poison but it's empty so you're okay but does that even make sense? If you knew it was poison why even try to drink it? Stroke of Luck you'll actually know - you get a stroke of luck and reach for this questionable phial that would have contained some really potent stuff but of course it's empty so you get nothing.

So basically:
0 roll & stroke of luck > Stroke of Luck message and Empty message
0 roll & curse of loki > Empty message (and you lose nothing)
0 roll > Empty message
Non-0 roll & stroke of luck > Stroke of Luck message and Grows stronger message
Non-0 roll && curse of loki > Poison message (and oops I accidentally left in the Grows stronger message)
Non-0 roll > Grows stronger message

helpers/levels.js
```diff
@@ -11,6 +11,50 @@ const getLevel = (xp = 0, prevPrevThreshold = 0, prevThreshold = 50, level = STA
 	return getLevel(xp, prevThreshold, threshold, level + 1);
 };
 
+const describeLevels = (...levels) => {
+	let level;
+	let compare;
+	let description;
+	let difference;
```
cormacmccarthy: `level` -> `opponentLevel`
`compare` -> `playerLevel`

helpers/levels.js
```diff
@@ -11,6 +11,50 @@ const getLevel = (xp = 0, prevPrevThreshold = 0, prevThreshold = 50, level = STA
...
+
+	if (levels.length === 1) {
+		level = levels[0];
+		description = level ? `level ${level}` : 'beginner';
```
cormacmccarthy: You'll probably want to change this to:

```
playerLevel = levels[0];
description = playerLevel ? `level ${playerLevel}` : 'beginner';
```

or
```
description = levels[0] ? `level ${levels[0]}` : 'beginner';
```

if you make the changes I suggested above.

helpers/levels.js
```diff
@@ -11,6 +11,50 @@ const getLevel = (xp = 0, prevPrevThreshold = 0, prevThreshold = 50, level = STA
...
+		compare = levels.shift();
+		level = Math.round(levels.reduce((total, lvl) => total + lvl, 0) / levels.length);
+		difference = compare - level;
+
+		description = `average level of ${level || 'beginner'}`;
+	}
+
+	return {
+		level,
+		compare,
```
cormacmccarthy: These two are returned but are never used that I can see. Furthermore, as is written now, "level" seems to be overloaded to be either the player's level, the single opponent's level, or the average of all opponent's levels. The changes suggested above will make these more consistent I think.

PR #212, opened 2018-01-23
Fix berserk

cormacmccarthy:
The fatigue calculation was all messed up, this fixes that.

Also renames "diminish" -> "fatigue" because that makes more sense.

Adds a bunch of unit tests, removes a redundant unit test.

Tried to make things more "functional" :)

### DIFF
--- a/cards/berserk.js
+++ b/cards/berserk.js
@@ -84,7 +84,7 @@ Stroke of luck increases damage per hit by 1.`;
 	}
 
 	increaseFatigue (currentFatigue) {
-		let fatigueAmount = (currentFatigue >= 0) ? currentFatigue : this.baseFatigue;
+		const fatigueAmount = (currentFatigue > this.baseFatigue) ? currentFatigue : this.baseFatigue;
 
 		return fatigueAmount + 1;
 	}
--- a/cards/berserk.js
+++ b/cards/berserk.js
@@ -82,15 +82,15 @@ Stroke of luck increases damage per hit by 1.`;
 		return roll({ primaryDice: this.attackDice, modifier, bonusDice: player.bonusAttackDice, crit: true });
 	}
 
-	fatigueIntBonus (currentBonus, iteration) {
-		let newBonus = currentBonus;
+	fatigueIntBonus (currentBonusFatigue, iteration) {
+		let newBonusFatigue = currentBonusFatigue;
 
 		if (iteration > 1) {
-			newBonus = (newBonus >= 0) ? newBonus : this.baseDiminishment;
-			newBonus += 1;
+			newBonusFatigue = (newBonusFatigue >= 0) ? newBonusFatigue : this.baseDiminishment;
+			newBonusFatigue += 1;
 		}
 
-		return newBonus;
+		return newBonusFatigue;
 	}
 
 	effectLoop (iteration, player, target, ring, activeContestants) {
--- a/cards/berserk.js
+++ b/cards/berserk.js
@@ -29,7 +29,7 @@ class BerserkCard extends HitCard {
 		this.damageAmount = this.options.damage;
 		this.iterations = 0;
 		this.baseDiminishment = -1;
-		this.intBonusDiminishment = undefined;
+		this.intBonusFatigue = undefined;
 	}
 
 	set bigFirstHit (bigFirstHit) {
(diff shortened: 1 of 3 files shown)


cards/berserk.js
```diff
@@ -63,21 +66,45 @@ Stroke of luck increases damage per hit by 1.`;
...
+		return modifier;
+	}
+
+	getAttackRoll (player, target) {
+		// once you hit the first time, you get a fatiguing bonus for each subsequent hit.
+		const modifier = this.getAttackRollBonus(player, target);
 		return roll({ primaryDice: this.attackDice, modifier, bonusDice: player.bonusAttackDice, crit: true });
 	}
 
+	fatigueIntBonus (currentBonus, iteration) {
```
tweettypography: Do we really have something called `intBonusFatigue` _and_ something called `fatigueIntBonus`? Feels confusing.

cards/berserk.js
```diff
@@ -63,21 +66,45 @@ Stroke of luck increases damage per hit by 1.`;
 		return roll({ primaryDice: this.damageDice });
 	}
 
-	getAttackRoll (player) {
-		// once you hit the first time, you get a diminishing bonus for each subsequent hit.
-		const modifier = player.dexModifier + (this.iterations > 1) ? Math.max(player.intModifier - this.iterations, 0) : 0;
+	getAttackRollBonus (player) {
+		let modifier = player.dexModifier;
+
+		if (this.iterations > 1) {
```
tweettypography: Why are we checking `iterations` here but doing anything with them within? Seems like we should either be using `iterations` or checking `intBonusFatigue`

cards/berserk.js
```diff
@@ -63,21 +66,45 @@ Stroke of luck increases damage per hit by 1.`;
...
 		return roll({ primaryDice: this.attackDice, modifier, bonusDice: player.bonusAttackDice, crit: true });
 	}
 
+	fatigueIntBonus (currentBonus, iteration) {
+		let newBonus = currentBonus;
+
+		if (iteration > 1) {
+			newBonus = (newBonus >= 0) ? newBonus : this.baseDiminishment;
+			newBonus += 1;
+		}
```
tweettypography: I don't really understand this logic. Wasn't the point to fatigue the bonuses? So if the iterations are not greater than 1 we just keep our existing bonus. If they _are_ greater than 1 and our current bonus is greater than or equal to zero then we just keep adding to it. If our current bonus is less than 0 then we set it to the value of baseDiminishment.

tweettypography: Maybe you just mean this to be `newBonus -= 1;`? I could see that making sense. Otherwise I'm missing something.

cards/berserk.js
```diff
@@ -63,21 +66,44 @@ Stroke of luck increases damage per hit by 1.`;
...
+	}
+
+	getAttackRoll (player, target) {
+		// once you hit the first time, you get a fatiguing bonus for each subsequent hit.
+		const modifier = this.getAttackRollBonus(player, target);
 		return roll({ primaryDice: this.attackDice, modifier, bonusDice: player.bonusAttackDice, crit: true });
 	}
 
+	increaseFatigue (currentFatigue) {
+		let fatigueAmount = (currentFatigue >= 0) ? currentFatigue : this.baseFatigue;
```
tweettypography: What about `const fatigueAmount = (currentFatigue > this.baseFatigue) ? currentFatigue : this.baseFatigue;`

cards/berserk.js
```diff
@@ -63,21 +66,44 @@ Stroke of luck increases damage per hit by 1.`;
...
+
+		return fatigueAmount + 1;
+	}
+
 	effectLoop (iteration, player, target, ring, activeContestants) {
 		this.iterations = iteration;
+		// intBonus doesn't kick in until we've actually successfully hit, don't fatigue the bonus until then
+		if (iteration > 1) {
+			this.intBonusFatigue = this.increaseFatigue(this.intBonusFatigue);
+		}
```
tweettypography: Or, we could just do:

```js
increaseFatigue () {
  this.intBonusFatigue += 1;
}

resetFatigue () {
  this.intBonusFatigue = 0;
}

resetCard () {
...
  this.resetFatigue();
}

...

if (iteration > 2) this.increaseFatigue();

if (strokeOfLuck) this.resetFatigue();
```

PR #218, opened 2018-01-24
Allow battles with multiple bosses

tweettypography:
Allows multiple bosses to be in the room at the same time, and won't start the fight if there are no human players in the ring even if there are multiple bosses. Bosses also won't target each other by default.

### DIFF
--- a/helpers/targeting-strategies.js
+++ b/helpers/targeting-strategies.js
@@ -2,6 +2,7 @@ const sample = require('lodash.sample');
 
 const TARGET_HIGHEST_HP_PLAYER = 'TARGET_HIGHEST_HP_PLAYER';
 const TARGET_HIGHEST_XP_PLAYER = 'TARGET_HIGHEST_XP_PLAYER';
+const TARGET_HUMAN_PLAYER_WEAK = 'TARGET_HUMAN_PLAYER_WEAK';
 const TARGET_LOWEST_HP_PLAYER = 'TARGET_LOWEST_HP_PLAYER';
 const TARGET_MAX_HP_PLAYER = 'TARGET_MAX_HP_PLAYER';
 const TARGET_NEXT_PLAYER = 'TARGET_NEXT_PLAYER';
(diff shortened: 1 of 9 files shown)


ring/index.js
```diff
@@ -219,29 +220,51 @@ class Ring extends BaseClass {
...
+					hasBoss = true;
+					return false;
+				}
+
+				return true;
+			});
+
+			return {
+				contestants: playerContestants,
+				numberOfMonstersInRing: playerContestants.length + (hasBoss ? 1 : 0)
```
cormacmccarthy: Does this calculate correctly if there is more than one boss in the ring?

helpers/targeting-strategies.js
```diff
@@ -44,6 +45,20 @@ function getTarget ({ playerContestant, contestants = [], strategy = TARGET_NEXT
 				return potentialTarget;
 			}, defaultTarget);
 		}
+		case TARGET_HUMAN_PLAYER_WEAK: {
```
cormacmccarthy: This seems to just be targeting a random human player rather than the weakest.

Also, if it truly isn't random and is picking a target, I'd rather see this target the human with the highest AC (the "tank") by default, and then for special boss battles we could have different strategies.

tweettypography: Not "weak" human player, it's the "weak" version of target human player - just does the one next to you or looks for a random one. The "strong" version would probably allow bosses to win collectively and to try to take out key players.

ring/index.js
tweettypography: Yes - basically if there are 1 or 100 bosses in the ring they're only "worth" 1

helpers/targeting-strategies.js
tweettypography: I guess I could just call it `TARGET_HUMAN_PLAYER_RANDOM` if you like that better, though it will use the "player next to me" strategy over random if possible so it would be a bit of a misnomer.

PR #220, opened 2018-01-25
Initial items PR

tweettypography:
Allow hydration of items and transfer of items between creatures (in particular, between a character and their monsters). Also adds a lottery ticket as an example of something that can be used by a character rather than a monster.

### DIFF
--- a/cards/sandstorm.js
+++ b/cards/sandstorm.js
@@ -79,7 +79,8 @@ class SandstormCard extends BlastCard {
 							effectResult: `${this.icon} lost in a cloud of blinding sand kicked up by`,
 							player: sandstormPlayer,
 							target: sandstormTarget,
-							ring
+							ring,
+							narration: `In the confusion of the sandstorm, ${sandstormTarget.givenName} will have trouble targeting ${sandstormTarget.pronouns.his} action.`
 						});
 
 						const oldTargets = getTargets.call(card, player, proposedTarget, ring, activeContestants);
--- a/cards/sandstorm.js
+++ b/cards/sandstorm.js
@@ -80,7 +80,7 @@ class SandstormCard extends BlastCard {
 							player: sandstormPlayer,
 							target: sandstormTarget,
 							ring,
-							narration: `In the confusion of the sandstorm, ${sandstormTarget.givenName} will have trouble targeting ${sandstormTarget.pronouns.his} action.`
+							narration: `In the confusion of the sandstorm, ${sandstormTarget.givenName} will have trouble with ${sandstormTarget.pronouns.his} aim.`
 						});
 
 						const oldTargets = getTargets.call(card, player, proposedTarget, ring, activeContestants);
--- a/game.js
+++ b/game.js
@@ -247,36 +247,15 @@ class Game extends BaseClass {
 						.catch(err => log(err));
 				},
 				equipMonster ({ monsterName, cardSelection } = {}) {
-					let selectedCards;
-
-					if (cardSelection) {
-						selectedCards = getArray(cardSelection);
-						if (selectedCards && !Array.isArray(selectedCards)) selectedCards = [selectedCards];
-					}
-
-					return character.equipMonster({ monsterName, cardSelection: selectedCards, channel })
+					return character.equipMonster({ monsterName, cardSelection: getArray(cardSelection), channel })
 						.catch(err => log(err));
 				},
 				giveItemsToMonster ({ monsterName, itemSelection } = {}) {
-					let selectedItems;
-
-					if (itemSelection) {
-						selectedItems = getArray(itemSelection);
-						if (selectedItems && !Array.isArray(selectedItems)) selectedItems = [selectedItems];
-					}
-
-					return character.giveItemsToMonster({ monsterName, itemSelection: selectedItems, channel })
+					return character.giveItemsToMonster({ monsterName, itemSelection: getArray(itemSelection), channel })
 						.catch(err => log(err));
 				},
 				takeItemsFromMonster ({ monsterName, itemSelection } = {}) {
-					let selectedItems;
-
-					if (itemSelection) {
-						selectedItems = getArray(itemSelection);
-						if (selectedItems && !Array.isArray(selectedItems)) selectedItems = [selectedItems];
-					}
-
-					return character.takeItemsFromMonster({ monsterName, itemSelection: selectedItems, channel })
+					return character.takeItemsFromMonster({ monsterName, itemSelection: getArray(itemSelection), channel })
 						.catch(err => log(err));
 				},
 				callMonsterOutOfTheRing ({ monsterName } = '') {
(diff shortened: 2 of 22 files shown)


cards/sandstorm.js
```diff
@@ -79,7 +79,8 @@ class SandstormCard extends BlastCard {
 							effectResult: `${this.icon} lost in a cloud of blinding sand kicked up by`,
 							player: sandstormPlayer,
 							target: sandstormTarget,
-							ring
+							ring,
+							narration: `In the confusion of the sandstorm, ${sandstormTarget.givenName} will have trouble targeting ${sandstormTarget.pronouns.his} action.`
```
cormacmccarthy: Sounds almost like they are "targeting their action" ("aiming _at_ their action"). There's got to be a better phrasing for this, but I'm having trouble finding it. Maybe... "aiming at targets"?

game.js
```diff
@@ -246,16 +247,15 @@ class Game extends BaseClass {
...
-					if (cardSelection) {
-						if (cardSelection.includes(', ')) {
-							selectedCards = cardSelection.split(', ');
-						} else {
-							selectedCards = cardSelection.split(' ');
-						}
-					}
-
-					return character.equipMonster({ monsterName, cardSelection: selectedCards, channel })
+					return character.equipMonster({ monsterName, cardSelection: getArray(cardSelection), channel })
```
cormacmccarthy: I think it would be great if you had to give cards to monsters rather than being able to give the same card to multiple monsters.

We'd need a fast take-back-all-cards command though... Otherwise it would get really really tedious to move the cards around between monsters.

Anyways, I won't complain if you add it to the PR but I certainly won't hold the PR up for it.

helpers/get-array.mocha.node.js
```diff
@@ -0,0 +1,54 @@
...
+
+		expect(getArray(input)).to.deep.equal(output);
+	});
+
+	it('can get an array from comma-and-space-separated numbers', () => {
+		const input = '1, 2';
+		const output = ['1', '2'];
+
+		expect(getArray(input)).to.deep.equal(output);
+	});
```
cormacmccarthy: Would love to see it be able to get an array from _space_ separated numbers so you didn't have to put in commas at all.

items/helpers/transfer.js
```diff
@@ -0,0 +1,102 @@
...
+const { sortItemsAlphabetically } = require('./sort');
+const chooseItems = require('./choose');
+
+module.exports = ({ from, to, itemSelection, channel }) => {
+	const checkEncounter = (arg) => {
+		if (to.inEncounter) {
+			return Promise.reject(channel({
+				announce: `You cannot give items to ${to.givenName} while they are fighting!`
+			}));
+		}
```
cormacmccarthy: So, this keeps you from giving a potion to a monster in battle, correct? Would you just use the potion yourself and it would apply to the monster then?

items/helpers/transfer.mocha.node.js
```diff
@@ -0,0 +1,179 @@
...
+		clock = sinon.useFakeTimers();
+		channelStub.resolves();
+	});
+
+	afterEach(() => {
+		clock.restore();
+		channelStub.reset();
+	});
+
+	it('can give a card to a monster', () => {
```
cormacmccarthy: 'a card' -> 'an item'

cormacmccarthy: Not that I wouldn't love to see this apply to cards as well... I got a little falsely excited when I saw this unit test :)

cards/sandstorm.js
tweettypography: "finding his target"?

game.js
tweettypography: I'm rearranging equip slightly in this PR but it's mostly about items so that will have to wait for now

helpers/get-array.mocha.node.js
tweettypography: That already works! I added a test just now to demonstrate

items/scrolls/lottery-ticket.js
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	action (character) {
+		const ticketNumber = random(1000, 1999);
+
+		this.emit('narration', {
+			narration: `🤞 ${character.givenName} holds a ticket imprinted with the numbers "${ticketNumber}".`
+		});
+
+		const winningNumber = random(1000, 1999);
```
cormacmccarthy: Oh man, I _really_ like this. I might actually go in someday and update this to act more like an actual lottery ticket with more properly weighted odds. If I understand correctly, a lottery would behave somewhat like the following:

```
const ticketNumbers = [random(1, 9), random(1, 9), random(1, 9), random(1, 9), random(1, 9)];
const winningNumbers = [random(1, 9), random(1, 9), random(1, 9), random(1, 9), random(1, 9)];

const winnings = ticketNumbers.reduce((winnings, number) => winningNumbers.find(number) ? winnings * 10 || 10 : winnings, 0);
```

items/helpers/transfer.js
tweettypography: Not 100% sure how it will work yet but in general yes, a character can use an item on a monster or can use an item without a monster if that item doesn't require one. When you give an item to a monster it means they are holding it and it _may_ affect them automatically in some way. The thing that's restricted here is just the giving and taking of items from someone who is in an encounter, not the using.

items/helpers/transfer.mocha.node.js
tweettypography: Fixed

cards/sandstorm.js
cormacmccarthy: Yeah, I thought about that but that's not exactly right because if they were just trying to find their target, they wouldn't "fire" unless they did find them. I think they're "firing" no matter what, so it's more the aiming that they are having the problem with.

items/scrolls/lottery-ticket.js
tweettypography: Nice! Something like that. I think your current method would actually give a bit too much away

monsters/base.js
```diff
@@ -40,6 +41,10 @@ class BaseMonster extends BaseCreature {
 		});
 	}
 
+	get itemSlots () { // eslint-disable-line class-methods-use-this
+		return DEFAULT_ITEM_SLOTS;
```
cormacmccarthy: May want to add a hook here for items. Probably can wait, but imagine that we have a "bat-belt" or "bag-of-holding" item that gives you extra slots... I won't hold the PR up for it though.

cards/sandstorm.js
tweettypography: how about just "aiming" then?

monsters/base.js
tweettypography: Good point, though we'll actually need hooks in a bunch of places once we get there so it might be better to wait and do that all at once so that we can come up with the best system

cards/sandstorm.js
tweettypography: Or "with his aim"?

cormacmccarthy: Oh, yeah, I like that. "with his aim"

items/scrolls/lottery-ticket.js
cormacmccarthy: Yeah, I was initially going match specific indices but didn't know how to do that with reduce off the top of my head :P 

There, edited it to be less generous :)

PR #221, opened 2018-01-25
take XinWey's Doctrine into account

cormacmccarthy:
This pr:

- Fixes cumulativeDamage tracker in berserk
- Allows BattleFocus to permakill player _if_ the target has only ~1HP and the player successfully damages them for _max damage_.
- Allows players to be hit after they are dead (they still are not target-able after they are dead, so the attack hitting them after they are dead _has to be_ a combination attack which is begun when they are alive, eg Forked Metal Rod, Horn Gore, BattleFocus, Berserk)
- Updates permadeath announcement to include XinWey's Doctrine (which is from a Brandon Sanderson sci-fi/fantasy novella)

### DIFF
--- a/cards/battle-focus.mocha.node.js
+++ b/cards/battle-focus.mocha.node.js
@@ -7,6 +7,12 @@ const pause = require('../helpers/pause');
 
 const { GLADIATOR } = require('../helpers/creature-types');
 
+const ultimateComboNarration = [];
+for (let i = 17; i < 101; i++) {
+	ultimateComboNarration.push(`HUMILIATION! ${i} hits`);
+}
+ultimateComboNarration.push('ULTIMATE COMBO! 100 HITS (99 total damage).');
+
 describe('./cards/battle-focus.js', () => {
 	let channelStub;
 	let pauseStub;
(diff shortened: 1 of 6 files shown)


cards/battle-focus.mocha.node.js
```diff
@@ -7,6 +7,12 @@ const pause = require('../helpers/pause');
 
 const { GLADIATOR } = require('../helpers/creature-types');
 
+const ultimateComboNarration = [];
+for (let i = 17; i < 101; i++) {
+	ultimateComboNarration.push(`HUMILIATION! ${i} hits`);
+}
+ultimateComboNarration.push('ULTIMATE COMBO! 100 HITS (99 total damage).');
```
tweettypography: I'm not sure I understand how you can hardcode 99 total damage here?

tweettypography: Oh because this is the test haha. I didn't get that at first

creatures/base.js
```diff
@@ -521,11 +518,11 @@ Battles won: ${this.battles.wins}`;
 			prevHp: originalHP
 		});
 
-		if (hp <= 0) {
+		if (!this.dead && hp <= 0) {
 			return this.die(assailant);
```
tweettypography: Don't we use this to track who killed who?

cormacmccarthy: crap. I did not realize there wasn't an attribute which was set to true/false by the .die method. I'm going to have to re-work this.

image

cormacmccarthy: Fixed.

PR #227, opened 2018-01-26
Delayed hit

cormacmccarthy:


cards/delayed-hit.js
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	get stats () {
+		return `Delay your turn. Use the delayed turn to immediately hit the next player who hits you.
+${super.stats}`;
+	}
+
+	effect (delayingPlayer, target) { // eslint-disable-line no-unused-vars
+		const when = Date.now();
+		const EFFECT_TYPE = `DelayedHitEffect${when}${random(0, 999999)}`;// Make sure these can stack, without the random sometimes in testing two would end up with the same EFFECT_TYPE
```
tweettypography: You don't have to do this! Just use `this.instanceId` if you need to know that the same card was used. It's a symbol and will always be unique to a single card. That said, you probably don't need either one. You can simply do `delayingPlayer.encounterEffects.filter(encounterEffect => encounterEffect !== delayedHitEffect)` as you only need to reference an effectType if you _want_ to catch all of the events of the same type (not allowing events to stack, for example) rather than this specific instance.

cards/delayed-hit.js
```diff
@@ -0,0 +1,64 @@
...
+
+		const delayedHitEffect = ({
+			card,
+			phase,
+			ring
+		}) => {
+			if (phase === DEFENSE_PHASE) {
+				const lastHitByOther = delayingPlayer.encounterModifiers.hitLog.find(hitter => hitter.assailant !== delayingPlayer);
+				if (lastHitByOther.when > when) {
```
tweettypography: Hm.. This might not work exactly as you expect. a plays the card > b hits a > the cards activates against b during c's turn. Is that correct? Or a plays the card > b hits a > a plays another card > the card activates against b then b plays again.

cards/delayed-hit.js
```diff
@@ -0,0 +1,64 @@
...
+		};
+
+		delayedHitEffect.effectType = EFFECT_TYPE;
+		delayingPlayer.encounterEffects = [...delayingPlayer.encounterEffects, delayedHitEffect];
+	}
+}
+
+DelayedHit.cardType = 'Delayed Hit';
+DelayedHit.probability = UNCOMMON;
+DelayedHit.description = 'Delay your turn, and to attack the next player who hits you.';
```
tweettypography: "and to" doesn't make sense

cards/helpers/hydrate.js
```diff
@@ -18,6 +20,14 @@ const hydrateDeck = (deckJSON = [], monster) => {
 	deck = deck.map(cardObj => hydrateCard(cardObj, monster));
 	deck = sortCardsAlphabetically(deck);
 
+	// give them two delayedHit cards if they don't have them already.
+	// To be ripped out as soon as everyone has their 2
+	const delayedHitCard = new DelayedHitCard();
+	if (!deck.find(card => card.cardType === delayedHitCard.cardType)) {
+		deck.push(new DelayedHitCard());
+		deck.push(new DelayedHitCard());
+	}
```
tweettypography: This looks like it should work, we just need to make sure we get it out right away, but if it was working I'd expect the hydration test to fail right here: https://github.com/deck-monsters/deck-monsters/blob/master/characters/helpers/hydrate.mocha.node.js#L34

cards/delayed-hit.js
cormacmccarthy: 😈 😮 and 🤡 are all in a match, playing in the order listed.

😈's turn. 😈 plays delayedHit card
😮's turn. 😮 plays hit card. 😮 hits 😈 
🤡's turn. 😈's delayed hit triggers and hits 😮, 🤡 then takes their turn as normal.

I had to hook into the pre-turn encounterEffects because there was no turn-cleanup-phase to hook in to, and this was basically the same difference. Eventually I could definitely see expanding turns to include:

GLOBAL_RING_EFFECTS
PRE_TURN_ATTACK_PHASE
PRE_TURN_DEFENSE_PHASE
PLAY
POST_TURN_CLEANUP_PHASE

(note that I put global ring effects at the top. Not sure why they happen after player specific attack/defense phases, but nothing hooks into them right now).

We would then move this card's resolution to the POST_TURN_CLEANUP_PHASE.

tweettypography: Okay, as long as that's the behavior you wanted I'm okay with it. And I'm going to do a few tweaks to effect logic soon so if there are things you'd like to see feel free to create issues and assign them to me.

cards/helpers/constants.js
```diff
@@ -1,3 +1,3 @@
 module.exports = {
-	DEFAULT_MINIMUM_CARDS: 14
+	DEFAULT_MINIMUM_CARDS: 20
```
tweettypography: That's a pretty big jump - but I guess it will help out players who are just getting started so I'm okay with it.

cards/helpers/hydrate.js
cormacmccarthy: It did! Updated unit tests to account for this.

cards/helpers/constants.js
cormacmccarthy: Well, we give them a preset 15 card deck at this point, so it really just gives them 5 random cards. What I'm not sure about is if their monsters have to be able to hold those cards, or if they are just _any_ random 5 cards.

cards/helpers/hydrate.js
tweettypography: Ah, I see that you fixed that test below but GHE had it collapsed. Not sure if you had done that before my first review or not but looks good.

tweettypography:
Pulled down your branch and the game crashed the first time I ran it... No errors either, probably because you aren't putting your effect into the promise chain.

screenshot 2018-01-26 12 05 26

tweettypography:
I'm also not sure that character played the card twice or if the card's effect fired twice.

PR #234, opened 2018-01-28
Messaging tweaks

cormacmccarthy:
Moves closer to having cards report all of their mechanics.

### DIFF
--- a/cards/brain-drain.mocha.node.js
+++ b/cards/brain-drain.mocha.node.js
@@ -29,7 +29,8 @@ describe('./cards/brain-drain.js', () => {
 		const hit = new HitCard({ damageDice: '1d4' });
 
 		const stats = `${hit.stats}
-Curse: xp -20`;
+Curse: xp -20
+(up to a maximum total of pre-battle XP - 40)`;
 
 		expect(brainDrain).to.be.an.instanceof(BrainDrainCard);
 		expect(brainDrain.icon).to.equal('🤡');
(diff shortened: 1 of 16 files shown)


cards/brain-drain.mocha.node.js
```diff
@@ -29,7 +29,8 @@ describe('./cards/brain-drain.js', () => {
 		const hit = new HitCard({ damageDice: '1d4' });
 
 		const stats = `${hit.stats}
-Curse: xp -20`;
+Curse: xp -20
+(up to a maximum total of pre-battle XP - 40)`;
```
tweettypography: This is a little bit confusing. The point is that you can get down as low as 40 xp and then you stop draining. So what you wrote is correct but maybe there's a better way to explain?

cards/cloak-of-invisibility.js
```diff
@@ -22,6 +22,11 @@ class CloakOfInvisibilityCard extends BaseCard {
 		super({ icon });
 	}
 
+	get stats () { // eslint-disable-line class-methods-use-this
+		return `You are invisible until you play a card that targets another player, or for the next 2 turns (whichever comes first).
+1d20 vs your int for opponent to see you on their turn (natural 20 removes your cloak).`;
```
tweettypography: Should we mention that currently you can stay invisible for more than 2 turns if you're immobilized / time shifted?

PR #238, opened 2018-01-29
Lay the groundwork for team battles

tweettypography:
Switches cards to use the targeting strategies and changes the strategies so that they take will take teams into account once we have those.

Also adds a description of current targeting strategy to monster stats.


tweettypography:
Merged as part of #240

PR #243, opened 2018-01-31
Rework immobilize cards

cormacmccarthy:
Immobilize now has 3 stages:

1. Initial Immobilize
2. Attack (hit)
3. Break Free

All bonuses etc are now correctly applied during all phases.

Each phase is independently over-write-able.

Stats are now fleshed out and mechanics accurately reported.


battlefield.js
```diff
@@ -76,6 +105,17 @@ return Promise
...
+		charCards = [
+			new EntranceCard(),
+			new EntranceCard(),
+			new EntranceCard(),
+			new EntranceCard(),
+			new EntranceCard(),
+			new EntranceCard(),
+			new EntranceCard(),
+			new EntranceCard(),
+			new EntranceCard()
```
tweettypography: We should set the battlefield back to normal before merging

cormacmccarthy: Agreed, that's one of the last things I have to do still...

cards/immobilize.js
```diff
@@ -86,47 +80,45 @@ class ImmobilizeCard extends HitCard {
...
+			advantageModifier = '- disadvantage';
+		} else if (strongBonus >= 0 && weakBonus < 0) {
+			advantageModifier = '+/- advantage/disadvantage';
 		}
 
-		return `${super.stats}${strModifiers}`;
+		return `If already immobilized, hit instead.
+${super.stats}
+${strModifiers}
+Opponent breaks free by rolling 1d20 vs immobilizer's ${this.freedomSavingThrowTargetAttr.toUpperCase()} ${advantageModifier} - (turns immobilized * 3)
```
tweettypography: We mostly display these in lowercase

cards/immobilize.js
```diff
@@ -139,145 +131,188 @@ class ImmobilizeCard extends HitCard {
...
+
+						card.play = () => Promise.resolve(true);
 					}
+				} else {
+					target.encounterEffects = target.encounterEffects.filter(effect => effect.effectType !== 'ImmobilizeEffect');
 
-					return card;
-				};
+					this.emit('narration', {
+						narration: `${target.givenName} is no longer ${this.actions.IMMOBILIZED}. ${target.pronouns.he} pushes the limp dead body of ${player.givenName} off of ${target.pronouns.him}self and proudly stands prepared to fight`
```
tweettypography: Should be `${capitalize(target.pronouns.he)}`

cards/immobilize.js
cormacmccarthy: That's true. I have it on the back-burner to go through and change that everywhere while I also get rid of the terms "spell bonus" "attack bonus" etc in favor of "INT bonus", "STR bonus" etc to more coherently tie the bonus back to the attr that granted it. I went back and forth on having it be lower case in this PR and fixing it in that one... If you want me to make it lower case here for consistency until I do that PR that's fine.

tweettypography: I still think it should be “int bonus” even then. “ac, hp, int” vs “AC, HP, INT” - it’s a little cleaner and easier to read when a battle is underway

cards/immobilize.js
```diff
@@ -13,30 +13,28 @@ const { signedNumber } = require('../helpers/signed-number');
...
 		icon = '😵',
+		freedomSavingThrowTargetAttr,
 		freedomThresholdModifier,
 		ongoingDamage,
 		strongAgainstCreatureTypes,
-		weakAgainstCreatureTypes,
+		targetProp,
 		uselessAgainstCreatureTypes,
+		weakAgainstCreatureTypes,
 		...rest
```
tweettypography: Why are so many of these things set as part of the options rather than as static properties?

cards/immobilize.js
```diff
@@ -139,145 +131,188 @@ class ImmobilizeCard extends HitCard {
...
+						target.encounterModifiers.immobilizedTurns = (target.encounterModifiers.immobilizedTurns || 0) + 1;
+						if (this.ongoingDamage > 0) {
 							this.emit('narration', {
-								narration: `${target.givenName} is no longer ${this.actions[2]}. ${capitalize(target.pronouns.he)} pushes the limp dead body of ${player.givenName} off of ${target.pronouns.him}self and proudly stands prepared to fight`
+								narration: `${target.givenName} takes ongoing damage from being ${this.actions.IMMOBILIZED}`
 							});
+							target.hit(this.ongoingDamage, player, this);
 						}
+
+						card.play = () => Promise.resolve(true);
```
tweettypography: We don’t actually use this anymore but _technically_ you should be returning whether or not the target is dead, not just always returning true.

cards/immobilize.js
```diff
@@ -139,145 +131,188 @@ class ImmobilizeCard extends HitCard {
...
+		const canHaveEffect = !this.uselessAgainstCreatureTypes.includes(target.creatureType);
+
+		if (alreadyImmobilized || !canHaveEffect) {
 			let narration = '';
 			if (alreadyImmobilized) {
-				narration = `${target.givenName} is already ${this.actions[2]}, now _show no mercy_!`;
+				narration = `${target.givenName} is already immobilized, now _show no mercy_!`;// Use immobilize here, because it could be the result of ANY immobilization, not just "coil" or whatever is checking right now.
 			} else {
-				narration = `${target.givenName} laughs hautily as you try to ${this.actions[0]} them, vent your fury at their mockery!`;
+				narration = `${target.givenName} laughs hautily as you try to ${this.actions.IMMOBILIZE} them, vent your fury at their mockery!`;
```
tweettypography: I think this should be haughtily? Also we usually use the monster’s name or pro Lund rather than “you”

cards/immobilize.js
cormacmccarthy: I guess I'm just used to seeing stats all caps, or at least initial caps. Kind of makes them stand out a little more as Things rather than just words. I'd rather go with consistency either way though and I don't want this PR to be even bigger. Maybe we'll hash this out again "some day".

PR #244, opened 2018-02-01
Updates player handbook, monster manual, and DMG. Also, makes cards r…

cormacmccarthy:
…eport all of their stats when not in battle.

### DIFF
--- a/dungeon-master-guide.js
+++ b/dungeon-master-guide.js
@@ -3,17 +3,17 @@ const Promise = require('bluebird');
 
 const { actionCard, itemCard } = require('./helpers/card');
 
-const BaseClass = require('./shared/baseClass');
-
 const allCards = require('./cards/helpers/all.js');
-const cardList = allCards.map(Card => `${new Card().cardType}`);
 const allItems = require('./items/helpers/all.js');
-const itemList = allItems.map(Item => `${new Item().itemType}`);
+
+const cardList = allCards.map(({ cardType }) => cardType);
+const itemList = allItems.map(({ itemType }) => itemType);
 
 const dungeonMasterGuide = (channel) => {
 	const { channelManager, channelName } = channel;
 
-	const header = `
+	const header =
+`
 \`\`\`
 				██████╗ ███╗   ███╗ ██████╗
 				██╔══██╗████╗ ████║██╔════╝
@@ -22,9 +22,9 @@ const dungeonMasterGuide = (channel) => {
 				██████╔╝██║ ╚═╝ ██║╚██████╔╝
 				╚═════╝ ╚═╝     ╚═╝ ╚═════╝
 
-	*The Card Catalogue:*
+*The Card Catalogue:*
 
-	${cardList.join(', ')}
+${cardList.join(', ')}
 \`\`\`
 `;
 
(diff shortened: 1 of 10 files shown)


dungeon-master-guide.js
```diff
@@ -0,0 +1,43 @@
+/* eslint-disable class-methods-use-this, max-len */
+const { actionCard, itemCard } = require('./helpers/card');
+
+const BaseClass = require('./shared/baseClass');
+
+const allCards = require('./cards/helpers/all.js');
+const allItems = require('./items/helpers/all.js');
+const cardCatalogue = allCards.reduce((catalogue, Card) => catalogue + `${actionCard(new Card(), true)}\n\n`, '');
+const itemCatalogue = allItems.reduce((catalogue, Card) => catalogue + `${itemCard(new Card(), true)}\n\n`, '');
```
tweettypography: Take a look at https://github.com/deck-monsters/deck-monsters/blob/master/creatures/base.js#L530

Basically, you need to send a series of messages to the queue so that they can be intelligently split up rather than that concatenating a giant string. The reason being that Slack has API message limits.

I'll also note that this doesn't feel like it needs to be a class - we mostly uses classes when we need saved state, events, and/or inheritance (usually all three) whereas this looks like it's just a function. Open to arguments either way though!

monster-manual.js
```diff
@@ -0,0 +1,45 @@
...
+const { monsterCard } = require('./helpers/card');
+
+const BaseClass = require('./shared/baseClass');
+
+const allMonsters = require('./monsters/helpers/all.js');
+const monsterCatalogue = allMonsters.reduce((catalogue, Monster) => catalogue + `${monsterCard(new Monster(), true)}\n\n`, '');
+const monsterList = allMonsters.reduce((list, Monster) => list + `${new Monster().creatureType}\n`, '');
+
+
+class MonsterManual extends BaseClass {
```
tweettypography: Same comments apply here re: class and message splitting as do with the DMG

player-handbook.js
```diff
@@ -38,23 +38,64 @@ class PlayerHandbook extends BaseClass {
...
+    \`look at [monster name]\` - Look at specified monster (yours or another player's)
+
+    *The Ring*
+
+    The glorious crucible of your monsters destiny.
+
+    \`send monster to ring\` - Select a monster and send it to the ring
 
-    \`send [monster name] to the ring\` - Send provided monster to the ring
+    \`send [monster name] to ring\` - Send provided monster to the ring
```
tweettypography: "the" is allowed here - it's what I always type. Just an fyi

player-handbook.js
```diff
@@ -38,23 +38,64 @@ class PlayerHandbook extends BaseClass {
...
+
+    \`look at [item name]\` - Look at specified item
+
+    \`use item\` - Use one of your items on yourself
+
+    \`give item to [monster name]\` - Give an item to one of your monsters
+
+    \`take item from [monster name]\` - Take an item from one of your monsters
+
+    \`use item on [monster name]\` - Use one of your items on yourself
```
tweettypography: Should we make a comment that any of these can be run as `use <item name> on <monster name>` or any combination of `an item` / `a monster` if you want to use the wizards

tweettypography:
Looks good!

player-handbook.js
cormacmccarthy: I agree, which leads to confusion when you type "jane look at the ring"... If we ever get _that_ sorted out, we should definitely put the "the" back here. Until then, I think it should remain an easter egg for consistency.

tweettypography: I'll fix it next in Jane when I deploy all these changes.

tweettypography: Change is pushed and ready to go out with the next deploy.

PR #252, opened 2018-02-03
Game display changes

cormacmccarthy:
Terser cards
Announce rounds, turns
attack/spell/hit bonus -> str/int/dex bonus


card-catalogue.js
```diff
@@ -0,0 +1,42 @@
...
+				|C.--. ||A.--. ||R.--. ||D.--. ||S.--. |
+				| :/\\: || (\\/) || :(): || :/\\: || :/\\: |
+				| :\\/: || :\\/: || ()() || (__) || :\\/: |
+				| '--'C|| '--'A|| '--'R|| '--'D|| '--'S|
+				\`------'\`------'\`------'\`------'\`------'
+
+*The Card Catalogue:*
+
+${cardList.join('\n')}
+\`\`\`
```
tweettypography: It occurs to me that there probably isn’t a reason to have these lines (or the similar ones below for items) wrapped in a code block. The cards themselves and artwork of course should have them.

PR #263, opened 2018-02-06
Fix turns

tweettypography:
Also adds some improved messaging to Sandstorm


cormacmccarthy:
I don't have time to pull down and test right now, but from what I saw when I reviewed the code, it looked good.

PR #267, opened 2018-02-07
Tweaks boost mechanics to mirror curse more closely

cormacmccarthy:
When boosting a prop, if you hit the max, the boost applies to HP instead (usually boosts are only 1 or 2 points, so this is not very much for you to heal, but better than nothing).

If the boost is applied to your AC, when you get hit with a melee attack (must be melee so people can't just use Blast and Sandstorm to clear people's ac boosts), the damage gets soaked up by your AC first, and then starts coming out of your HP. This is a subtle difference but will mean two things:

1. You can now safely use more than 1 or 2 thick-skin cards (or the like) and not be penalized by your boost cap.
2. Using 4 thick-skin cards in a row will boost you against hits only until you actually get hit. At which point, your HP will still be protected from the damage by your boost soaking up as much as it can, but then your boost is gone and you have to fill it back up again.

So, this kind of nerfs boosts in one regard, while also _boosting_ (har har) them in another. And crap I just realized that this doesn't apply to Adrenaline Rush or Ecdysis because they don't extend boost... 🤔 ok well, once this gets merged I'll make another PR "some day" that marries Boost with Ecdysis (I should have just fixed boost to begin with rather than making an entirely new boost card with ecdysis... oh well ⌛️ )

I honestly can not remember why I've been sitting on this code for almost a month. I feel like it was either that I was working on some closely related changes in another branch (which I still am) and wanted to wait until that was ready and push these at the same time, or that we discussed it and someone expressed dissatisfaction with this mechanic or something... I just can't remember. So, I'm going to go ahead and submit this for PR and see what the response is :)

The boost narrative is all mostly Fallout inspired FWIW.


cards/boost.js
```diff
@@ -29,8 +31,72 @@ class BoostCard extends BaseCard {
...
+	}
+
+	getBoostOverflowNarrative (player, target) {
+		return `${target.givenName}'s ${this.cursedProp} boosts have been maxed out. Boost will be granted to hp instead.`;
+	}
+
 	effect (player, target) {
-		target.setModifier(this.boostedProp, this.boostAmount);
+		const preBoostedPropValue = target[this.boostedProp];
+		let boostAmount = Math.abs(this.boostAmount);
```
tweettypography: Why `Math.abs`?

creatures/base.js
```diff
@@ -572,23 +572,43 @@ Battles won: ${this.battles.wins}`;
 		});
 		this.encounterModifiers.hitLog = hitLog;
 
-		const hp = this.hp - damage;
-		const originalHP = this.hp;
+		if (this.encounterModifiers.ac >= damage) {
+			this.encounterModifiers.ac -= damage;
 
-		this.hp = hp;
+			this.emit('narration', {
```
tweettypography: I don't think we have a `creature.narration` listener yet (like we do for items and cards) so you'd need to add one if you want to use this.

tweettypography:
Did you run `lint:fix` on this? Also, have you tested in battlefield? I'd be surprised if the narration was actually being output.

cards/boost.js
cormacmccarthy: Can't remember now, but it's no longer necessary. Removed.

PR #271, opened 2018-02-11
Adds cardClass to all cards 

cormacmccarthy:
This is to help distinguish between AOE, psychic, melee, etc.

This will immediately be used by changes to Boost so that AC boosts only soak up melee attacks (that way people can't use Blast and Sandstorm to get to clear AC boosts "for free"), but this also replaces the old isAreaOfEffect boolean.


cards/blast.js
```diff
@@ -44,13 +44,13 @@ class BlastCard extends BaseCard {
 	}
 }
 
+BlastCard.cardClass = 'AOE';
```
tweettypography: The classes should really come from a standard set of constants rather than hard coded strings on the cards. Also, could there ever be a need for two? As in, should this accept an array?

cards/prion-disease.js
```diff
@@ -113,12 +113,12 @@ Usually restores between 0-3hp to each opponent, and 1-4hp for the player.
 	}
 }
 
+PrionDiseaseCard.cardClass = 'Poison';
 PrionDiseaseCard.cardType = '1993-09-7202 18:58';
 PrionDiseaseCard.probability = EPIC.probability;
 PrionDiseaseCard.description = 'Buy a questionable round of milkshakes for everyone.';
 PrionDiseaseCard.level = 2;
 PrionDiseaseCard.cost = EXPENSIVE.cost;
-PrionDiseaseCard.isAreaOfEffect = true;
```
tweettypography: The reason this was AOE was so it wouldn’t be affected by cloaks, faceswap, etc. Does poison work the same way? This is actually a good example of one that should maybe be an array :)

cormacmccarthy: My thought was that it is hard to serve someone a drink who you can't see... But I could see for this one card making it both AOE and Poison.

tweettypography: I mean, that’s definitely true. But these are regular milkshakes and I like the idea that everyone drinks one, no exception.

PR #272, opened 2018-02-12
[HOLD] Gh12   exploration   find cards

cmcculloh-kr:
Changes to Jane's deck-monsters.js script:
```
	const initializeGame = () => {
		if (!game) {
			const log = msg => robot.logger.info(msg);
			const savedState = robot.brain.get('deckMonster.savedState') || oldState;
			const mainRing = { channelName: 'MAIN_RING', channel: channel({ id: constants.HUBOT_DECK_MONSTERS_RING }) };
			const theWorld = { channelName: 'THE_WORLD', channel: channel({ id: constants.HUBOT_DECK_MONSTERS_WORLD }) };

			if (savedState) {
				try {
					game = restoreGame([mainRing, theWorld], savedState, log);
				} catch (ex) {
					robot.logger.error('deck-monsters:', ex);

					mainRing({
						announce: 'Failed to restore the saved game state.'
					});

					game = new Game([mainRing, theWorld], {}, log);

					// TODO: Ideally we should use a fail-safe instead wher we store an index to a saved state and if it fails we generate a new index.
					// That way there's no chance of losing the old saved state. As it stands right now if this failed because the savedState wasn't
					// properly retrieved we'd still have lost the original state (because it would be overwritten).
					robot.brain.set(`deckMonster.savedErrorState.${Date.now()}`, savedState); // Save a backup because something just went very wrong
				}
			} else {
				game = new Game([mainRing, theWorld], {}, log);
			}

			game.saveState = state => robot.brain.set('deckMonster.savedState', state);
		}
	};
```


tweettypography:
Let's reverse how channels work - the game should ask the adapter for a channel rather than the adapter passing in a list of channels.

PR #278, opened 2018-02-20
Changes how psychic cards are handled

tweettypography:
Adds a concussion card to go with brain drain. Also changes the targeting of enthrall & entrance (which will impact how it's affected by Sandstorm, etc). Finally, changes the way that cloaks / camo handle psychic cards - the caster will have to roll to see through the invisibility but _won't_ look elsewhere for a target.

### DIFF
--- a/cards/cloak-of-invisibility.js
+++ b/cards/cloak-of-invisibility.js
@@ -7,7 +7,7 @@ const { roll } = require('../helpers/chance');
 const { BARD, CLERIC, WIZARD } = require('../constants/creature-classes');
 const { ATTACK_PHASE, DEFENSE_PHASE } = require('../constants/phases');
 const { capitalize } = require('../helpers/capitalize');
-const { AOE, HIDE, PSYCHIC } = require('../constants/card-classes');
+const { AOE, HIDE, MUSICAL, PSYCHIC } = require('../constants/card-classes');
 const { RARE } = require('../helpers/probabilities');
 const { PRICEY } = require('../helpers/costs');
 
@@ -55,7 +55,8 @@ class CloakOfInvisibilityCard extends BaseCard {
 				card.effect = (player, target, ring, activeContestants) => {
 					if (phase === DEFENSE_PHASE &&
 						player !== invisibilityTarget && target === invisibilityTarget &&
-						!card.isCardClass(AOE)) {
+						!card.isCardClass(AOE) &&
+						!card.isCardClass(MUSICAL)) {
 						// If this isn't a psychic attack and your target is invisible, look for a new target
 						if (!card.isCardClass(PSYCHIC)) {
 							const potentialTargets = activeContestants.filter(({ monster }) => (monster !== player && !isInvisible(monster)));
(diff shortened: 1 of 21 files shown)


cards/concussion.mocha.node.js
```diff
@@ -0,0 +1,81 @@
...
+		const hit = new HitCard({ damageDice: '1d4' });
+
+		const stats = `${hit.stats}
+Curse: xp -20
+can reduce xp down to 40, then takes 4 from hp instead.`;
+
+		expect(concussion).to.be.an.instanceof(ConcussionCard);
+		expect(concussion.icon).to.equal('🥊');
+		expect(concussion.curseAmount).to.equal(-20);
+		expect(concussion.cursedProp).to.equal('xp');
```
cormacmccarthy: Can we make this target INT instead of XP? It will make it really useful _against_ clerics who rely heavily on INT and right now there is no way to curse INT other than just trying to reduce XP to bring down the INT bonus.

tweettypography: It was meant to be the pair to brain drain, but I do like that. I'll change

cards/cloak-of-invisibility.js
```diff
@@ -7,7 +7,7 @@ const { roll } = require('../helpers/chance');
 const { BARD, CLERIC, WIZARD } = require('../constants/creature-classes');
 const { ATTACK_PHASE, DEFENSE_PHASE } = require('../constants/phases');
 const { capitalize } = require('../helpers/capitalize');
-const { AOE, HIDE, PSYCHIC } = require('../constants/card-classes');
+const { AOE, HIDE, MUSICAL, PSYCHIC } = require('../constants/card-classes');
```
RodEsp: Can we make this acoustic so it can be more general? I can see there being cards like "Roar" that might not be musical per se but will still involve sound as the main medium.

cards/cloak-of-invisibility.js
```diff
@@ -56,17 +56,20 @@ class CloakOfInvisibilityCard extends BaseCard {
 					if (phase === DEFENSE_PHASE &&
 						player !== invisibilityTarget && target === invisibilityTarget &&
 						!card.isCardClass(AOE) &&
-						!card.isCardClass(PSYCHIC)) {
-						const potentialTargets = activeContestants.filter(({ monster }) => (monster !== player && !isInvisible(monster)));
+						!card.isCardClass(MUSICAL)) {
+						// If this isn't a psychic attack and your target is invisible, look for a new target
+						if (!card.isCardClass(PSYCHIC)) {
```
RodEsp: Should this line be indented further in?

characters/helpers/random.js
```diff
@@ -68,6 +68,13 @@ module.exports = ({
 		let deck = cleanBossDeck(getMinimumDeck());
 		deck = cleanBossDeck(fillDeck(deck, {}, character));
 		character.deck = fillDeck(deck, {}, character);
+
+		// Also level up some Kalevalas
```
RodEsp: Noooooooo, they don't need any help! Lol

cards/cloak-of-invisibility.js
cormacmccarthy: Yeah, something like Acoustic or Aural might be better.

cards/cloak-of-invisibility.js
cormacmccarthy: I had the same confusion, but no it shouldn't. It just looks weird because of all the `&&` conditions off the original if.

cards/cloak-of-invisibility.js
tweettypography: Sure thing

cards/cloak-of-invisibility.js
tweettypography: No, just looks weird because of multiline conditions above

PR #279, opened 2018-02-21
Add command handlers to the repo

tweettypography:
Makes the adapters a lot "dumber" and lighter-weight.


tweettypography:
Yes, I'd like to get some tests in for sure. Ideally there should be at least one for each command in the end.

tweettypography:
I'd also like to refactor eventually because at the moment this still uses the `getCharacter` in the main game file that acts as a sort of command list which we don't really need any more as that logic could move directly into these command files. I didn't remove it the moment because we need it for backwards compatibility.

PR #280, opened 2018-02-21
Adds ring summary

iowillhoit:
Shows a ring summary with `look at ring summary`

screenshot 2018-02-21 10 22 34

To test:
- Pull changes
- Copy over to local jane (manually or with `npm link`)
- Start jane
- Send monsters to the ring
- Run `dm look at ring summary`

### DIFF
--- a/ring/index.js
+++ b/ring/index.js
@@ -215,10 +215,9 @@ Sent the following monster into the ring:
 				});
 			}))
 			.then(() => {
-				const summary = this.contestants.map((contestant) => {
-					const { monster } = contestant;
-					return `${monster.identity} - ${monster.creatureType} (${monster.level})`;
-				});
+				const summary = this.contestants.map(({ monster }) => (
+					`${monster.identity} - ${monster.creatureType} (${monster.level})`
+				));
 
 				return channelManager.queueMessage({
 					announce:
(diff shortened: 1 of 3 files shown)


tweettypography:
This looks good. I don't mind that it's always at the end but maybe you should add a command `look at ring summary` that prints _only_ this out, as requested by @RodEsp? With the command processing now in this repo it should be relatively easy to do. Let me know if you have any questions!

ring/index.js
```diff
@@ -214,6 +214,20 @@ Sent the following monster into the ring:
 					channelName
 				});
 			}))
+			.then(() => {
+				const summary = this.contestants.map(({ monster }) => (
+					`${monster.identity} - ${monster.creatureType} (${monster.level})`
```
cormacmccarthy: Make sure to use whatever prop displays "beginner" rather than "0" here.

iowillhoit: `displayLevel` does, but it also has the word `level` before each one. I can strip that off though. Or maybe there is another prop that doesn't have the prefix

tweettypography: Good point - I'd love a general helper that takes numbered levels and can return appropriate formats, to be used everywhere (including in that prop). Sort of related, I also want to create one that lets you pass in two monsters and returns pronouns / names for the first with "himself" etc if it's the same monster because we have that logic all over

PR #307, opened 2026-04-09
Add local simulation harness (@deck-monsters/harness)

tweettypography:
## Summary

Adds **`@deck-monsters/harness`** for local ring simulations and **fixes the primary cause of out-of-order public ring narration**: concurrent `game.command` actions on the same `Game` from different users.

### Fix: per-room engine queue

- `createKeyedPromiseQueue()` in `@deck-monsters/engine` — serializes async work by key.
- `RoomManager.runSerializedEngineWork(roomId, fn)` wraps every `game.command` action.
- **Per-user `activeFlows` is unchanged** — still prevents one user from stacking commands during prompts.

### Harness

- Scenarios: `ring-two-bosses`, `concurrent-look-monsters` (mirrors production serialization via `createRoomCommandRunner`).
- Timer / dispose cleanup for reliable script exit.

### Docs

- `docs/harness-ring-feed-investigation.md` — issue context, root cause, shipped fix, reconnect gap note (still open).

### Tests

- Engine: `room-engine-queue.test.ts`
- Server: integration test uses `createRoomCommandRunner` + parallel `runCommand`
- Harness: mocha covers both scenarios

<div><a href="https://cursor.com/agents/bc-73ec4689-a2a8-4d70-976c-817e56a8b9e7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-73ec4689-a2a8-4d70-976c-817e56a8b9e7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


tweettypography (commented):
Good overall shape — the investigation doc is excellent context, consolidating the test helpers into the engine package is the right call, and the dispose cleanup (exploration timer, creature heal/respawn intervals, `Ring.clearRing`) was genuinely needed. One real bug to fix and a few smaller notes below.

---

### Bug: `Object.assign` mutates the shared event object

**`packages/harness/src/public-feed.ts` line 26**

```ts
events.push(Object.assign(event, { seq: seq++ }));
```

`Object.assign(event, …)` mutates the **original** event object that was handed to this `deliver` callback. The `RoomEventBus` does synchronous fan-out, so the same object reference may be held by other subscribers (e.g. the WebSocket ring feed, auto-responders). After this line every other handler that still holds a reference to that event will silently see an extra `seq` property on it.

Fix: create a new object instead of mutating:

```ts
events.push({ ...event, seq: seq++ });
```

---

### Silent behavioral change in `createTestGame` — `spawnBosses: false`

**`packages/engine/src/testing/index.ts` line 42**  
**`packages/server/src/shared/test-helpers.ts` (re-export)**

The new canonical `createTestGame` defaults to `spawnBosses: false`, whereas the old server version called `new Game({ roomId })` with no such option. Since `packages/server/src/shared/test-helpers.ts` now re-exports from the engine, every existing server integration test silently picks up this default.

This is probably the right default (boss timers were causing exit issues), but worth a quick audit of the server test suite to confirm nothing was relying on boss-spawning behaviour before merging. If the tests are already green in CI that's confirmation enough.

---

### Minor: seq assertion in harness test is trivially true

**`packages/harness/src/harness.test.ts` lines 20–23**

```ts
for (let i = 0; i < feed.events.length; i++) {
    expect(feed.events[i]!.seq).to.equal(i);
}
```

`seq` is assigned by a local counter inside `capturePublicFeed` — it's `0, 1, 2, …` by construction, so `events[i].seq === i` is always true regardless of publication order. Not wrong as a smoke test, but it doesn't actually validate the event ordering that the harness was built to study. A future test could assert on semantic ordering (e.g. round/turn headers arrive before the cards they describe).

---

### Minor: type casts could be avoided

**`packages/engine/src/game.ts`** and **`packages/engine/src/ring/index.ts`**

```ts
// game.ts
if (character && typeof character.disposeTimers === 'function') { … }
const monsters = character?.monsters as Array<{ disposeTimers?: () => void }> | undefined;

// ring/index.ts
(monster as { disposeTimers?: () => void })?.disposeTimers?.();
```

`disposeTimers` is now a public method on `BaseCreature`. If the static types for `character` and `monster` in these loops already extend `BaseCreature` (or have a `disposeTimers` property), the casts and runtime `typeof` checks are unnecessary. If the types genuinely don't surface the method, that's a type-hierarchy gap worth a follow-up (not a blocker here).


---

PR #318, opened 2026-04-11
Fix ring feed crashes, ordering, and live UX

tweettypography:
## Summary
- fix server ring `add` listeners to parse BaseClass emitter args safely, eliminating `contestant.isBoss` undefined crashes
- improve Ring pane UX by keeping auto-follow while the user is at bottom and hiding internal `ring.fightResolved` analytics events from player-facing feed rows
- fix ring history replay ordering after reload by serializing event persistence writes and deriving timestamps from eventId prefixes when available
- add monster-name autocomplete expansion for placeholder commands (e.g. `send [monster] to the ring`, `revive [monster]`) using the current user’s monster list
- persist user-submitted console commands as private history events and render them as input rows on reload so users see both what they sent and what the game responded with
- resolve PR review comments by removing duplicate `autocompleteMonsters` endpoint and memoizing monster-name arrays passed into command autocomplete
- stabilize `SurvivalKnifeCard` low-health test by stubbing heal outcome randomness so it no longer flakes on poison/zero-heal rolls

## Validation
- `pnpm --filter @deck-monsters/web build`
- `pnpm --filter @deck-monsters/web test -- src/__tests__/console-history-event-map.test.ts src/__tests__/useCommandAutocomplete.test.ts`
- `pnpm --filter @deck-monsters/server test -- src/integration/command-flow.test.ts src/room-manager.test.ts`
- `pnpm lint`
- `pnpm --filter @deck-monsters/engine exec mocha src/cards/survival-knife.test.ts`
- `for i in $(seq 1 30); do pnpm --filter @deck-monsters/engine exec mocha src/cards/survival-knife.test.ts >/tmp/survival_fixed_$i.log 2>&1 || { echo FAIL:$i; break; }; done; echo DONE`

<div><a href="https://cursor.com/agents/bc-034efe16-ad46-44a9-847f-4f1102c8d9db"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-034efe16-ad46-44a9-847f-4f1102c8d9db"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/apps/web/src/__tests__/console-history-event-map.test.ts
+++ b/apps/web/src/__tests__/console-history-event-map.test.ts
@@ -1,0 +1,34 @@
+import { describe, expect, it } from 'vitest';
+import { mapConsoleHistoryEvent } from '../utils/console-history-event-map.js';
+
+describe('mapConsoleHistoryEvent', () => {
+  it('maps console input system events to input rows', () => {
+    const mapped = mapConsoleHistoryEvent({
+      id: 'evt-1',
+      type: 'system',
+      text: 'send elm to the ring',
+      payload: { consoleInput: true },
+    });
+
+    expect(mapped).toEqual({
+      id: 'evt-1',
+      type: 'input',
+      text: 'send elm to the ring',
+    });
+  });
+
+  it('keeps regular system events as system rows', () => {
+    const mapped = mapConsoleHistoryEvent({
+      id: 'evt-2',
+      type: 'system',
+      text: '-- reconnecting --',
+      payload: {},
+    });
+
+    expect(mapped).toEqual({
+      id: 'evt-2',
+      type: 'system',
+      text: '-- reconnecting --',
+    });
+  });
+});
--- a/apps/web/src/__tests__/useCommandAutocomplete.test.ts
+++ b/apps/web/src/__tests__/useCommandAutocomplete.test.ts
@@ -42,4 +42,28 @@ describe('useCommandAutocomplete', () => {
       expect(equip.insertValue).not.toContain(']');
     }
   });
+
+  it('suggests monster-name command variants for send/revive placeholders', () => {
+    const { result } = renderHook(() => useCommandAutocomplete('send e', true, {
+      monsterNames: ['Elm', 'Westley'],
+      sendableMonsterNames: ['Elm'],
+      deadMonsterNames: ['Westley'],
+    }));
+    const labels = result.current.map(s => s.label.toLowerCase());
+    const inserts = result.current.map(s => s.insertValue.toLowerCase());
+
+    expect(labels).toContain('send elm to the ring');
+    expect(inserts).toContain('send elm to the ring');
+  });
+
+  it('suggests revive commands only when a dead monster name is available', () => {
+    const { result } = renderHook(() => useCommandAutocomplete('revive', true, {
+      monsterNames: ['Elm', 'Westley'],
+      sendableMonsterNames: ['Elm'],
+      deadMonsterNames: ['Westley'],
+    }));
+    const labels = result.current.map(s => s.label.toLowerCase());
+    expect(labels).toContain('revive westley');
+    expect(labels).not.toContain('revive elm');
+  });
 });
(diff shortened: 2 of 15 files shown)


tweettypography (commented):
Good overall direction — the serialized write queue, the `autoFollowRef` approach, and the utility-extraction refactors are all solid. Two issues need fixing before merge; a few smaller notes below.

---

## Must fix

### 1. Duplicate tRPC procedures (`router.ts`)

`myMonsters` and `autocompleteMonsters` are identical copy-pastes — same input schema, same body, same return shape. The client only calls `trpc.game.myMonsters`, so `autocompleteMonsters` is dead code. Please drop one of them.

```ts
// packages/server/src/trpc/router.ts  (~405-448)
myMonsters: protectedProcedure ...         // ← used by the web client
autocompleteMonsters: protectedProcedure ... // ← exact duplicate, nothing calls this
```

### 2. `useMemo` deps contain unstable array references (`ConsolePane.tsx` + `useCommandAutocomplete.ts`)

The arrays passed to `useCommandAutocomplete` are re-created on every render of `ConsolePane`:

```tsx
// ConsolePane.tsx
const suggestions = useCommandAutocomplete(
  inputValue,
  !activePromptId && !inputLocked,
  {
    monsterNames: monsterRows.map((m) => m.name),         // new [] every render
    deadMonsterNames: monsterRows.filter(...).map(...),    // new [] every render
    sendableMonsterNames: monsterRows.filter(...).map(...), // new [] every render
  }
);
```

The hook's `useMemo` deps are `[input, enabled, context.monsterNames, ...]` — but those arrays compare by reference, so the memo never caches across renders triggered by typing. Every keystroke recomputes the full suggestion list.

The simplest fix is to memoize the derived arrays in `ConsolePane`:

```tsx
const monsterNames = useMemo(() => monsterRows.map((m) => m.name), [monsterRows]);
const deadMonsterNames = useMemo(
  () => monsterRows.filter((m) => m.dead).map((m) => m.name),
  [monsterRows]
);
const sendableMonsterNames = useMemo(
  () => monsterRows.filter((m) => !m.dead && !m.inRing).map((m) => m.name),
  [monsterRows]
);

const suggestions = useCommandAutocomplete(inputValue, !activePromptId && !inputLocked, {
  monsterNames,
  deadMonsterNames,
  sendableMonsterNames,
});
```

Since `monsterRows` comes from a tRPC query (stable reference between fetches), these memos will only recompute when the monster list actually changes — not on every keystroke.

---

## Nits / low-priority

**`autoFollowRef` causes an extra state update on every live event** (`RingPane.tsx`)

```ts
useEffect(() => {
  if (autoFollowRef.current && feedRef.current) {
    feedRef.current.scrollTop = feedRef.current.scrollHeight;
    setIsAtBottom(true);  // ← fires a re-render even when already true
  }
}, [events]);
```

`setIsAtBottom(true)` when we're already at the bottom enqueues a render on every new ring event. Could guard it: `if (!isAtBottom) setIsAtBottom(true)`. Minor, but worth knowing.

**`timestampFromEventId` — no upper-bound sanity check** (`game-event-map.ts`)

A malformed `eventId` like `99999999999999-junk` produces a far-future timestamp that passes the `> 0` check. Not a real concern for well-formed IDs, but a `ts <= Date.now() + someBuffer` guard would prevent history ordering weirdness if data ever gets corrupted.

**`event-persister.ts` — `onError` can fire after cleanup**

The returned unsubscribe function stops new events from coming in but doesn't drain or cancel `writeQueue`. Inflight writes can still invoke `onError` after teardown. Probably acceptable; just worth a comment so future readers don't wonder.

---

## What's good

- Serialized write queue is the right fix for the ordering bug, and the test in `event-persister.test.ts` is well-constructed — verifying the single-in-flight constraint with deferred promises is exactly the right technique.
- Using `autoFollowRef` instead of state for the scroll guard avoids stale-closure bugs; this is a genuine improvement.
- Extracting `historyEventToConsoleEvent` → `mapConsoleHistoryEvent` into its own module makes it independently testable — good call.
- `shouldRenderRingEvent` with a `Set<string>` is clean and trivially extensible.
- `timestampFromEventId` fallback logic is correct and well-tested.


---

PR #324, opened 2026-04-11
Fix stuck console prompt flow recovery

tweettypography:
## Summary
- fix prompt-flow stickiness by restoring pending prompts server-side and client-side so active prompt answers are routed correctly
- harden prompt responses against stale request IDs: event bus now returns handled/not-handled, router returns `PRECONDITION_FAILED` for stale prompt answers, and the web console refetches and rehydrates the latest pending prompt snapshot when this happens
- reduce perceived "fight flew by" behavior caused by reconnect replay bursts by resuming ring/console subscriptions from the latest tracked event ID after transport errors instead of replaying recent buffered events from scratch
- add configurable delay pacing controls using midpoint env vars per delay type (`*_MIDPOINT_MS`) with derived min/max ranges while preserving existing defaults
- add optional round-based pacing shaping and optional per-delay caps (`*_CAP_MS`) without changing default behavior
- thread current round into ring turn delays so optional round shaping applies to live combat pacing
- add per-room turn-gap telemetry (`dm_turn_gap_ms`) from `card.played` events and reset boundaries across encounters

## Root cause notes
- production logs show rapid `respondToPrompt` bursts and intermittent feed reconnect/open/close cycles
- core ring pacing delays are active in engine; the "fast" feel aligned with replay bursts during reconnect plus relatively short baseline delay values
- equip intermittency aligned with stale prompt responses being silently no-oped before hard-fail + rehydration handling

## Validation
- `pnpm build`
- `pnpm --filter @deck-monsters/engine test -- --grep "delay-times|ring/index\.ts"`
- `pnpm --filter @deck-monsters/server test -- src/metrics/collector.test.ts`

## Files touched
- `packages/engine/src/events/room-event-bus.ts`
- `packages/engine/src/events/room-event-bus.test.ts`
- `packages/server/src/trpc/router.ts`
- `packages/server/src/trpc/router.test.ts`
- `apps/web/src/components/ConsolePane.tsx`
- `apps/web/src/components/RingPane.tsx`
- `packages/engine/src/helpers/delay-times.ts`
- `packages/engine/src/helpers/delay-times.test.ts`
- `packages/engine/src/ring/index.ts`
- `packages/server/src/metrics/index.ts`
- `packages/server/src/metrics/collector.ts`
- `packages/server/src/metrics/collector.test.ts`

<div><a href="https://cursor.com/agents/bc-c540bd99-1073-41d4-b8bf-0e78d95d5554"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-c540bd99-1073-41d4-b8bf-0e78d95d5554"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/apps/web/src/components/ConsolePane.tsx
+++ b/apps/web/src/components/ConsolePane.tsx
@@ -85,13 +85,13 @@ export default function ConsolePane({ roomId, isActive, onEvent }: ConsolePanePr
   const [quickActions, setQuickActions] = useState<QuickAction[]>([]);
   const [suggestionIndex, setSuggestionIndex] = useState(-1);
 
-  // Always undefined — the server delivers the last 100 events on connect and
-  // seenRef handles deduplication with DB history. Previously this was set from
-  // the history effect, but changing a subscription input restarts the WebSocket.
-  const subLastEventId: string | undefined = undefined;
+  // Resume cursor for reconnects. Updated only on errors so we don't restart
+  // the subscription on every event.
+  const [subLastEventId, setSubLastEventId] = useState<string | undefined>(undefined);
   const virtuosoRef = useRef<VirtuosoHandle>(null);
   const inputRef = useRef<HTMLInputElement>(null);
   const seenRef = useRef(new Set<string>());
+  const latestTrackedEventIdRef = useRef<string | undefined>(undefined);
   const historyApplied = useRef(false);
 
   // Register command-insert function so external callers (CommandReference, etc.) can populate the input
@@ -104,7 +104,7 @@ export default function ConsolePane({ roomId, isActive, onEvent }: ConsolePanePr
 
   // Fetch persistent console history from DB on mount
   const { data: history } = trpc.game.consoleHistory.useQuery({ roomId });
-  const { data: pendingPrompt } = trpc.game.pendingPrompt.useQuery(
+  const { data: pendingPrompt, refetch: refetchPendingPrompt } = trpc.game.pendingPrompt.useQuery(
     { roomId },
     { enabled: !!roomId },
   );
@@ -269,6 +269,8 @@ export default function ConsolePane({ roomId, isActive, onEvent }: ConsolePanePr
         // Keep-alive ping from server — no UI action needed
         if (event.type === 'heartbeat') return;
 
+        latestTrackedEventIdRef.current = tracked.id;
+
         if (seenRef.current.has(tracked.id)) return;
         seenRef.current.add(tracked.id);
 
@@ -373,6 +375,7 @@ export default function ConsolePane({ roomId, isActive, onEvent }: ConsolePanePr
           });
         }
         setReconnecting(true);
+        setSubLastEventId(latestTrackedEventIdRef.current);
       },
     }
   );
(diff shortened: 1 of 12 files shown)


tweettypography (commented):
Reviewed the full diff. Overall this is solid — the root-cause analysis in the PR description maps cleanly to each change, the test coverage is good, and the defaults are backward-compatible. Two issues found and fixed on `claude/review-pr-324-e3NpT` (same base, ready to merge into this branch or squash on top).

---

## Fixes applied

### 1. `hasPendingPrompt` removed — dead code (`room-event-bus.ts`)

`hasPendingPrompt(requestId)` was added but never called anywhere in the codebase. `respondToPrompt`'s boolean return value and `getPendingPromptForUser` already cover every real call-site. Removed to keep the public API surface clean.

### 2. Duplicate `latestTrackedEventIdRef.current` assignment hoisted in `RingPane.tsx`

The ref was being set in two places:
- once inside the `ring.state` early-return branch
- again on the next line after that branch

Each event path only hit one assignment, so there was no double-write, but the duplication was confusing. Hoisted the single assignment to before the `ring.state` check (matching the cleaner pattern already used in `ConsolePane.tsx`).

---

## Everything else checked out

**`delay-times.ts` defaults preserve existing ranges** — verified the midpoint math:
- `very_short` midpoint 1500 → `[1000, 2000]` ✓ (was `Math.ceil(random*1000)+1000`)
- `short` midpoint 2250 → `[1500, 3000]` ✓
- `medium` midpoint 3000 → `[2000, 4000]` ✓
- `long` midpoint 4500 → `[3000, 6000]` ✓

**`ring.fight` eventName handling in `collector.ts`** — confirmed the ring actually emits `{ eventName: 'fightBegins' }` (ring/index.ts:503) and `{ eventName: 'fightConcludes' }` (ring/index.ts:708), so the collector's reset and completion logic are wired to real events. The backward-compat fallback for missing `eventName` is also safe.

**`round` variable scope in `ring/index.ts`** — `let round = 1` is declared in `fight()` and incremented inside `doAction()`. All four `veryShortDelay(round)` / `shortDelay(round)` call-sites are within the `doAction` closure, so `round` is correctly in scope.

**Router `handleCommand` reorder** — moving the `activeFlows` gate before `handleCommand` is correct. A user mid-flow sees the "answer the current prompt first" message (with the pending prompt snapshot) regardless of whether the new command would have been recognized. Better UX than the old silent "not recognized" during an active flow.

**`PRECONDITION_FAILED` + client recovery** — the throw-and-refetch pattern is clean. The client catches the error, shows the message, then calls `refetchPendingPrompt()` to rehydrate the latest prompt from the server. If the prompt has since expired, `data` will be null and no stale prompt is injected. Race-condition-safe.

**`ConsolePane` scroll behavior** — `upsertPendingPrompt` appends to the end of `consoleEvents`; Virtuoso's `followOutput="smooth"` handles auto-scroll when the user is already at the bottom. No forced scroll needed.

**`collector.test.ts` gap count** — the comment "two gaps should be recorded: 1500ms and 2200ms" is accurate (encounter 1), and one more gap of 1250ms comes from encounter 2, totalling 3 as asserted.

---

PR #330, opened 2026-04-12
feat(engine): weighted boss level distribution by room monster levels

tweettypography:
## Summary
- implement weighted boss-level selection for ring boss spawns:
  - 20% keep the current fully random boss level distribution
  - 30% cap boss level at highest player monster level in ring + 1
  - 50% cap boss level at floor(average player monster level in ring)
- prioritize in-ring player monster levels for boss scaling; when ring has no player monsters, fall back to all player monster levels in the room
- keep boss spawn timing ring-focused (empty ring still uses beginner pacing), so room fallback affects level selection but not spawn cadence
- optimize level→XP cap lookup with exponential+binary search while still delegating level math to `getLevel()`
- ignore dead/destroyed room monsters in fallback level sourcing
- tighten roll announcement typing and behavior for partial roll payloads
- add tests for weighted boundaries, ring-priority/room-fallback source selection, ring-only timing context, dead/destroyed fallback filtering, partial roll announcements, and XP-cap boundary behavior

## Validation
- `pnpm --filter @deck-monsters/engine build && pnpm --filter @deck-monsters/engine test`
- `pnpm --filter @deck-monsters/engine test -- src/ring/index.test.ts`

<div><a href="https://cursor.com/agents/bc-a089943d-7732-454d-a855-eb4e538f860f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-a089943d-7732-454d-a855-eb4e538f860f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/packages/engine/src/announcements/rolled.ts
+++ b/packages/engine/src/announcements/rolled.ts
@@ -2,22 +2,27 @@ import { signedNumber } from '../helpers/signed-number.js';
 import type { RoomEventBus } from '../events/index.js';
 
 interface RollResult {
-	naturalRoll: { result: number };
-	bonusResult: number;
-	modifier: number;
+	naturalRoll?: { result?: number };
+	bonusResult?: number;
+	modifier?: number;
 	primaryDice?: string;
 	strokeOfLuck?: boolean;
 	curseOfLoki?: boolean;
-	result: number | string;
+	result?: number | string;
 }
 
 interface RolledOpts {
 	outcome?: string;
 	reason: string;
-	roll: RollResult;
+	roll?: RollResult;
 	vs?: number | string;
 	who: any;
 }
+
+const toNumber = (value: unknown, fallback = 0): number => {
+	const parsed = Number(value);
+	return Number.isFinite(parsed) ? parsed : fallback;
+};
 
 export function announceRolled(
 	eb: RoomEventBus,
@@ -25,19 +30,32 @@ export function announceRolled(
 	monster: any,
 	{ outcome, reason, roll, vs, who }: RolledOpts,
 ): void {
-	let rollDesc = `${roll.naturalRoll.result}${signedNumber(roll.bonusResult)}${signedNumber(roll.modifier)}`;
-	if (roll.primaryDice) rollDesc = `${rollDesc} on ${roll.primaryDice}`;
+	const naturalRoll = toNumber(roll?.naturalRoll?.result);
+	const bonusResult = toNumber(roll?.bonusResult);
+	const modifier = toNumber(roll?.modifier);
+	const fallbackResult = naturalRoll + bonusResult + modifier;
+	const result = roll?.result ?? fallbackResult;
+	const hasDetailedRollDescription = roll?.naturalRoll?.result !== undefined ||
+		roll?.bonusResult !== undefined ||
+		roll?.modifier !== undefined ||
+		!!roll?.primaryDice;
 
-	const text = `${who.givenName} rolled _${rollDesc}_ ${reason}`;
+	let rollDesc = hasDetailedRollDescription
+		? `${naturalRoll}${signedNumber(bonusResult)}${signedNumber(modifier)}`
+		: `${result}`;
+	if (roll?.primaryDice) rollDesc = `${rollDesc} on ${roll.primaryDice}`;
+
+	const whoName = who?.givenName ?? 'Someone';
+	const text = `${whoName} rolled _${rollDesc}_ ${reason}`;
 
 	const vsMsg = vs ? ` v ${vs}` : '';
-	let rollResult: string = roll.strokeOfLuck ? 'Nat 20!' : String(roll.result);
-	if (roll.curseOfLoki) rollResult = 'Crit Fail!';
+	let rollResult: string = roll?.strokeOfLuck ? 'Nat 20!' : String(result);
+	if (roll?.curseOfLoki) rollResult = 'Crit Fail!';
 
 	eb.publish({
 		type: 'announce',
 		scope: 'public',
 		text: `${text}\n🎲 *${rollResult}${vsMsg}*${outcome ? `\n    ${outcome}` : ''}\n `,
-		payload: { roll, who, outcome },
+		payload: { roll: roll ?? { result }, who, outcome },
 	});
 }
--- a/packages/engine/src/announcements/rolled.ts
+++ b/packages/engine/src/announcements/rolled.ts
@@ -1,15 +1,27 @@
 import { signedNumber } from '../helpers/signed-number.js';
 import type { RoomEventBus } from '../events/index.js';
 
-interface RollResult {
-	naturalRoll?: { result?: number };
-	bonusResult?: number;
-	modifier?: number;
+type RollFlags = {
 	primaryDice?: string;
 	strokeOfLuck?: boolean;
 	curseOfLoki?: boolean;
+};
+
+type RollWithParts = RollFlags & {
+	naturalRoll: { result: number };
+	bonusResult: number;
+	modifier: number;
 	result?: number | string;
-}
+};
+
+type RollWithResultOnly = RollFlags & {
+	result: number | string;
+	naturalRoll?: { result?: number };
+	bonusResult?: number;
+	modifier?: number;
+};
+
+type RollResult = RollWithParts | RollWithResultOnly;
 
 interface RolledOpts {
 	outcome?: string;
--- a/packages/engine/src/game.ts
+++ b/packages/engine/src/game.ts
@@ -75,7 +75,10 @@ export class Game extends BaseClass {
 
 		this.ring = new Ring(
 			this._eventBus,
-			{ spawnBosses: (this.options as any).spawnBosses },
+			{
+				spawnBosses: (this.options as any).spawnBosses,
+				getRoomMonsterLevels: () => this.getRoomMonsterLevels(),
+			},
 			this.log
 		);
 		this.exploration = new Exploration(this._eventBus as any, {}, this.log);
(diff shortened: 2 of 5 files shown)


packages/engine/src/ring/index.ts
```diff
@@ -20,8 +20,30 @@ const BOSS_SPAWN_MAX_DELAY_MS = 2_100_000; // 35 min
...
+	while (level <= targetLevel) {
+		const threshold = prevPrevThreshold + prevThreshold;
+		if (level === targetLevel) return threshold - 1;
+
+		prevPrevThreshold = prevThreshold;
+		prevThreshold = threshold;
+		level += 1;
+	}
+
+	return 49;
```
tweettypography: The fallback `return 49` at the bottom is unreachable dead code — the `while (level <= targetLevel)` loop always reaches `level === targetLevel` before exiting naturally, so it will always return inside the loop. Can be removed.

Also worth noting: this function silently mirrors the Fibonacci-like sequence inside `getLevel()` (from `helpers/levels.ts`) without reusing it. If the leveling formula ever changes in `getLevel`, `getXpCapForLevel` will silently drift and produce wrong XP caps. Consider computing this from `getLevel` directly:

```ts
const getXpCapForLevel = (targetLevel: number): number => {
    if (targetLevel <= 0) return 49;
    // find the first XP value at which getLevel returns targetLevel + 1, then subtract 1
    let xp = 0;
    while (getLevel(xp) <= targetLevel) xp++;
    return xp - 1;
};
```
Or extract the threshold computation from `getLevel` into a shared helper.

---

packages/engine/src/ring/index.ts
```diff
@@ -928,28 +969,36 @@ export class Ring extends BaseClass {
...
-			return randomContestant();
+	private determineBossLevelCap(playerLevels: number[], roll: number): number | undefined {
+		if (roll <= BOSS_FULL_RANDOM_WEIGHT_PERCENT) {
+			return undefined;
 		}
 
-		if (playerContestants.length === 0) {
-			// Same band as avg XP 0 when players are present: roughly level 0–1 bosses.
-			return randomContestant({ xp: random(0, 40) });
+		if (playerLevels.length <= 0) {
```
tweettypography, line 980: When the player levels array is empty and the roll lands in the "highest + 1" band (21–50), this returns `0` rather than `1` (i.e. it doesn't add the "+1"). That's consistent with the "treat as level 0 when nobody is home" intent, but it means the two capped bands are indistinguishable when the room is empty. A short comment here would make the design intent clear, e.g.:

```ts
// No monsters to scale from — collapse both capped bands to level 0.
if (playerLevels.length <= 0) {
    return 0;
}
```

---

packages/engine/src/ring/index.ts
```diff
@@ -918,7 +959,7 @@ export class Ring extends BaseClass {
 	 * no player monsters in the ring yet (empty ring still uses beginner boss pacing — not full random).
 	 */
 	private isBeginnerBossContext(): boolean {
-		return this.hasOnlyBeginnerPlayers() || this.getPlayerMonsterLevels().length === 0;
+		return this.hasOnlyBeginnerPlayers() || this.getPreferredBossScalingLevels().length === 0;
```
tweettypography: `hasOnlyBeginnerPlayers` and `isBeginnerBossContext` both now use `getPreferredBossScalingLevels()`, which falls back to room-level data when the ring is empty. This means room monster levels now influence boss spawn **timing** (12–22 min beginner schedule vs 20–35 min standard schedule), not just boss level caps.

Concretely: if a room has high-level monsters in characters but none in the ring, the boss timer will now fire on the standard (slower) schedule, whereas previously an empty ring always used the beginner (faster) schedule.

This seems intentional given the updated JSDoc comment on `isBeginnerBossContext`, but it's worth confirming — especially since the timing shift could surprise existing rooms that are mid-session after a deploy.

---

packages/engine/src/announcements/rolled.ts
```diff
@@ -2,42 +2,60 @@ import { signedNumber } from '../helpers/signed-number.js';
...
 	curseOfLoki?: boolean;
-	result: number | string;
+	result?: number | string;
 }
 
 interface RolledOpts {
 	outcome?: string;
 	reason: string;
-	roll: RollResult;
+	roll?: RollResult;
```
tweettypography: Making `naturalRoll`, `bonusResult`, `modifier`, and `result` all optional on `RollResult`, and making `roll` itself optional on `RolledOpts`, is quite broad. The stated motivation (heal-card stubs passing `{ result: 5 }`) only requires `result` to be optional and the components (`naturalRoll`, `bonusResult`, `modifier`) to be optional. `result` becoming optional too means a fully empty `roll: {}` is now valid and will silently produce `🎲 *0*`.

Consider a narrower approach: keep `result` required (or at least the natural-roll components required when `result` is absent) using a discriminated union:

```ts
type RollResult =
  | { result: number | string; naturalRoll?: { result?: number }; bonusResult?: number; modifier?: number; primaryDice?: string; strokeOfLuck?: boolean; curseOfLoki?: boolean }
  | { naturalRoll: { result: number }; bonusResult: number; modifier: number; result?: number | string; primaryDice?: string; strokeOfLuck?: boolean; curseOfLoki?: boolean };
```

That way callers must provide either a `result` or the full components, and the type system catches accidental empty objects.

---

packages/engine/src/game.ts
```diff
@@ -170,6 +173,15 @@ export class Game extends BaseClass {
...
 	}
 
+	private getRoomMonsterLevels(): number[] {
+		return Object.values(this.characters)
+			.flatMap((character: any) =>
+				Array.isArray(character?.monsters) ? character.monsters : []
+			)
+			.map((monster: any) => Number(monster?.level ?? 0))
+			.filter((level: number) => Number.isFinite(level) && level >= 0);
+	}
```
tweettypography, line 184: `getRoomMonsterLevels` includes every monster across all characters — including dead monsters and fresh level-0 placeholders. This could pull the computed average down significantly in rooms with many idle/dead low-level monsters, making bosses more beginner-friendly than intended.

If that's the design intent (use all monsters, not just active ones), that's fine — but it's worth a comment, or at least a filter to exclude dead/destroyed monsters:

```ts
.flatMap((character: any) =>
    Array.isArray(character?.monsters) ? character.monsters : []
)
.filter((monster: any) => !monster?.dead && !monster?.destroyed)
.map(...)
```

---

tweettypography (commented):
Overall the direction is solid — the weighted distribution is clean, the ring-priority / room-fallback logic is well-structured, and the test coverage of the new helper methods is good. Five things worth addressing before merging:

1. **`getXpCapForLevel` — unreachable fallback + Fibonacci drift risk** (see inline): the trailing `return 49` is dead code, and the function silently duplicates `getLevel()`'s sequence. If the leveling formula changes, XP caps will silently drift.

2. **Empty-room capped bands collapsing to 0 — needs a comment** (see inline): when no monsters are present, both the "highest+1" and "floor(average)" bands return `0`. That's fine as a design choice, but it's not obvious from the code.

3. **Room levels now affect boss spawn *timing*, not just level** (see inline): `hasOnlyBeginnerPlayers` and `isBeginnerBossContext` use `getPreferredBossScalingLevels()`, which falls back to room-level data. This silently changes the spawn timer schedule for rooms with high-level room monsters but no ring contestants.

4. **`RollResult` type is too permissive after going fully optional** (see inline): a discriminated union requiring either `result` or the component fields would give callers better type safety with no runtime cost.

5. **`getRoomMonsterLevels` includes dead/destroyed monsters** (see inline): worth confirming whether that's intentional, or if a `!monster.dead` filter should be added.

---

tweettypography (commented):
All five points from the first pass are cleanly addressed:

1. **`getXpCapForLevel`** — the Fibonacci reimplementation is gone; the new loop delegates to `getLevel()` directly. Verified manually: levels 0–5 produce the expected caps (49, 99, 149, 249, 399, 649), negative inputs clamp to 49, and floats floor correctly.

2. **Empty-room cap comment** — the inline comment makes the design intent unambiguous.

3. **Timing vs. level separation** — `isBeginnerBossTimingContext()` is ring-only now, which is the right split. Boss cadence is driven by who's actually fighting; boss difficulty is informed by the broader room context.

4. **`RollResult` union type** — `RollWithParts | RollWithResultOnly` correctly rejects an empty `{}` while accepting both the full-components form and the stub-friendly `{ result: N }` form. The optional chaining in `announceRolled` works correctly against both arms.

5. **Dead/destroyed monster filter** — the `!monster.dead && !monster.destroyed` guard in `getRoomMonsterLevels` is correct given that `dead` and `destroyed` are getters derived from `hp` in `BaseCreature`. The new test that verifies this by mutating `hp` directly is sound.

One minor observation (no action needed): the `getXpCapForLevel` loop is `O(maxXp)` rather than `O(level)`. For the Fibonacci-like sequence used here, a level-15 cap is ~49k XP, so the loop runs ~49k iterations for a high-level boss cap. Completely negligible given this fires at most once every 12–35 minutes, but worth a note if the function ever moves to a hot path.

LGTM — good to merge once the draft flag is lifted.

---

PR #331, opened 2026-04-12
Fix cross-room ring announcement bleed between active rooms

tweettypography:
## Summary
- guard `creature.levelUp` via `wrapGameEvent` so level-up announcements cannot leak across active rooms
- tighten the ownership guard with a typed `RoomScopedGame` shape, top-level fast-path checks, and a named max depth constant
- keep the previous card-play coverage and add explicit regressions for level-up and heal leakage across two concurrent rooms
- fix strict TypeScript typing in guard traversal to keep the engine build clean

## Testing
- `pnpm --filter @deck-monsters/engine build`
- `pnpm --filter @deck-monsters/engine exec mocha src/game.test.ts`

<div><a href="https://cursor.com/agents/bc-2ded0135-c63c-4b01-ab99-3ff3c2b4870b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-2ded0135-c63c-4b01-ab99-3ff3c2b4870b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/packages/engine/src/announcements/index.ts
+++ b/packages/engine/src/announcements/index.ts
@@ -75,23 +75,29 @@ function createRoomScopedEventGuard(game: RoomScopedGame): (...args: any[]) => b
 		if (value === game || value === game.ring || value === game.exploration) return true;
 
 		const characters = Object.values(game.characters ?? {});
-		for (const character of characters) {
+		for (const character of characters as Array<RoomScopedCharacter | undefined>) {
 			if (value === character) return true;
 
-			const monsters = Array.isArray(character?.monsters) ? character.monsters : [];
+			const monsters: unknown[] = Array.isArray(character?.monsters) ? character.monsters : [];
 			if (monsters.includes(value)) return true;
 
-			const charItems = Array.isArray(character?.items) ? character.items : [];
+			const charItems: unknown[] = Array.isArray(character?.items) ? character.items : [];
 			if (charItems.includes(value)) return true;
 
-			const deck = Array.isArray(character?.deck) ? character.deck : [];
+			const deck: unknown[] = Array.isArray(character?.deck) ? character.deck : [];
 			if (deck.includes(value)) return true;
 
 			for (const monster of monsters) {
-				const cards = Array.isArray(monster?.cards) ? monster.cards : [];
+				const cards: unknown[] =
+					monster && typeof monster === 'object' && Array.isArray((monster as any).cards)
+						? (monster as any).cards
+						: [];
 				if (cards.includes(value)) return true;
 
-				const items = Array.isArray(monster?.items) ? monster.items : [];
+				const items: unknown[] =
+					monster && typeof monster === 'object' && Array.isArray((monster as any).items)
+						? (monster as any).items
+						: [];
 				if (items.includes(value)) return true;
 			}
 		}
--- a/packages/engine/src/announcements/index.ts
+++ b/packages/engine/src/announcements/index.ts
@@ -49,12 +49,32 @@ import { announceXPGain } from './xpGain.js';
 import { announceLevelUp } from './level-up.js';
 import type { RoomEventBus } from '../events/index.js';
 
-function createRoomScopedEventGuard(game: any): (...args: any[]) => boolean {
+const MAX_OWNERSHIP_WALK_DEPTH = 3;
+
+type RoomScopedCharacter = {
+	monsters?: unknown[];
+	items?: unknown[];
+	deck?: unknown[];
+};
+
+type RoomScopedGame = {
+	eventBus: RoomEventBus;
+	ring: {
+		on: (event: string, listener: (...args: any[]) => void) => (...args: any[]) => void;
+		off: (event: string, listener: (...args: any[]) => void) => void;
+	};
+	exploration: unknown;
+	characters?: Record<string, RoomScopedCharacter | undefined>;
+	on: (event: string, listener: (...args: any[]) => void) => (...args: any[]) => void;
+	off: (event: string, listener: (...args: any[]) => void) => void;
+};
+
+function createRoomScopedEventGuard(game: RoomScopedGame): (...args: any[]) => boolean {
 	const ownsDirectly = (value: unknown): boolean => {
 		if (!value || typeof value !== 'object') return false;
 		if (value === game || value === game.ring || value === game.exploration) return true;
 
-		const characters = Object.values((game.characters ?? {}) as Record<string, any>);
+		const characters = Object.values(game.characters ?? {});
 		for (const character of characters) {
 			if (value === character) return true;
 
@@ -80,11 +100,14 @@ function createRoomScopedEventGuard(game: any): (...args: any[]) => boolean {
 	};
 
 	return (...args: any[]): boolean => {
+		// Hot path: the emitting instance is almost always a top-level argument.
+		if (args.some((arg) => ownsDirectly(arg))) return true;
+
 		const visited = new WeakSet<object>();
 
 		const walk = (value: unknown, depth: number): boolean => {
 			if (ownsDirectly(value)) return true;
-			if (depth >= 3 || !value || typeof value !== 'object') return false;
+			if (depth >= MAX_OWNERSHIP_WALK_DEPTH || !value || typeof value !== 'object') return false;
 
 			if (visited.has(value)) return false;
 			visited.add(value);
@@ -104,7 +127,7 @@ function createRoomScopedEventGuard(game: any): (...args: any[]) => boolean {
 	};
 }
 
-export function initialize(game: any): () => void {
+export function initialize(game: RoomScopedGame): () => void {
 	const eb: RoomEventBus = game.eventBus;
 	const isRoomScopedEvent = createRoomScopedEventGuard(game);
 
@@ -140,8 +163,10 @@ export function initialize(game: any): () => void {
 		{ event: 'potion.used', listener: wrapGameEvent(announceItemUsed) },
 		{
 			event: 'creature.levelUp',
-			listener: (_className: string, _instance: any, { monster, level }: { monster: any; level: number }) =>
-				announceLevelUp(eb, monster, level),
+			listener: wrapGameEvent(
+				(_eb, _className: string, _instance: any, { monster, level }: { monster: any; level: number }) =>
+					announceLevelUp(_eb, monster, level)
+			),
 		},
 		{ event: 'scroll.narration', listener: wrapGameEvent(announceNarration) },
 		{ event: 'scroll.used', listener: wrapGameEvent(announceItemUsed) },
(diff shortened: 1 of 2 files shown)


packages/engine/src/announcements/index.ts
```diff
@@ -49,53 +49,118 @@ import { announceXPGain } from './xpGain.js';
...
+		{ event: 'gainedXP', listener: wrapGameEvent(announceXPGain) },
+		{ event: 'item.narration', listener: wrapGameEvent(announceNarration) },
+		{ event: 'item.used', listener: wrapGameEvent(announceItemUsed) },
+		{ event: 'potion.narration', listener: wrapGameEvent(announceNarration) },
+		{ event: 'potion.used', listener: wrapGameEvent(announceItemUsed) },
 		{
 			event: 'creature.levelUp',
 			listener: (_className: string, _instance: any, { monster, level }: { monster: any; level: number }) =>
 				announceLevelUp(eb, monster, level),
 		},
```
tweettypography, line 176: **Bug: `creature.levelUp` is the only `gameEvents` entry still missing the ownership guard.**

This listener is registered via `game.on()`, which binds to the process-wide `globalSemaphore`, so a level-up fired in room A will still call `announceLevelUp` into room B's `eb`. All the neighbouring entries in `gameEvents` were correctly updated to `wrapGameEvent`, but this one was left as a bare lambda.

The second positional argument emitted by `BaseClass.emit` is always `this` (the levelling creature), so `isRoomScopedEvent` can catch it via `ownsDirectly`. A straightforward fix wraps the entire listener:

```ts
{
    event: 'creature.levelUp',
    listener: wrapGameEvent(
        (eb, _className, _instance, { monster, level }: { monster: any; level: number }) =>
            announceLevelUp(eb, monster, level)
    ),
},
```

Note that `wrapGameEvent` prepends `eb` then spreads the raw semaphore args, so the handler signature shifts by one compared to the bare lambda.

---

packages/engine/src/announcements/index.ts
```diff
@@ -49,53 +49,118 @@ import { announceXPGain } from './xpGain.js';
...
+				const cards = Array.isArray(monster?.cards) ? monster.cards : [];
+				if (cards.includes(value)) return true;
+
+				const items = Array.isArray(monster?.items) ? monster.items : [];
+				if (items.includes(value)) return true;
+			}
+		}
+
+		return false;
+	};
```
tweettypography, line 106: **Performance: `ownsDirectly` is an O(players × monsters × collections) linear scan that runs on every semaphore event.**

`Object.values(game.characters)` allocates a new array, then for each character iterates monsters, items, deck cards, and each monster's cards/items. In active multi-room combat this fires on `card.rolled`, `card.effect`, `creature.hit`, etc. — easily dozens of times per second per room. As the roster grows the cost compounds across all active listeners on the global semaphore.

A simpler, faster approach is to maintain a `Set<object>` of all owned references that's kept up-to-date as characters/monsters/cards are added or removed:

```ts
// built once per room, updated on roster changes
const ownedRefs = new Set<object>([game, game.ring, game.exploration]);

// rebuild when characters mutate (stateChange or explicit hook)
function rebuildOwnedRefs() {
    ownedRefs.clear();
    ownedRefs.add(game); ownedRefs.add(game.ring); ownedRefs.add(game.exploration);
    for (const char of Object.values(game.characters ?? {})) {
        ownedRefs.add(char);
        for (const m of char?.monsters ?? []) {
            ownedRefs.add(m);
            for (const c of m?.cards ?? []) ownedRefs.add(c);
            for (const i of m?.items ?? []) ownedRefs.add(i);
        }
    }
}
```

The guard then becomes a single `ownedRefs.has(arg)` check — O(1) per arg. The trade-off is wiring the rebuild to state mutations, but the `stateChange` event already exists for exactly this purpose.

---

packages/engine/src/announcements/index.ts
```diff
@@ -49,53 +49,118 @@ import { announceXPGain } from './xpGain.js';
...
+
+	return (...args: any[]): boolean => {
+		const visited = new WeakSet<object>();
+
+		const walk = (value: unknown, depth: number): boolean => {
+			if (ownsDirectly(value)) return true;
+			if (depth >= 3 || !value || typeof value !== 'object') return false;
```
tweettypography: **Minor: a new `WeakSet` and `walk` closure are allocated on every event call.**

```ts
return (...args: any[]): boolean => {
    const visited = new WeakSet<object>(); // allocated per invocation
    const walk = (value, depth) => { ... }; // closure created per invocation
```

For most combat events `ownsDirectly` will match on the first or second top-level arg (the emitting instance is always `args[1]` via `BaseClass.emit`), so `walk` is rarely needed beyond depth 0. A quick win: call `ownsDirectly` on all top-level args first, and only fall back to the deep-walk + `WeakSet` if none match. That avoids the allocation on the hot path.

Also, `3` on line 87 is a bare magic number. A named constant with a short comment (e.g. `MAX_WALK_DEPTH = 3 // card → payload wrapper → event envelope`) would make the intent clearer and the value easier to revisit.

---

packages/engine/src/game.test.ts
```diff
@@ -56,6 +57,42 @@ describe('game.ts', () => {
...
+			monster.cards = [card];
+
+			const character = new Beastmaster({ name: 'Room A Trainer' });
+			character.addMonster(monster);
+			roomA.characters = { ...roomA.characters, 'room-a-user': character };
+
+			globalSemaphore.emit('card.played', card.name, card, { player: monster });
+
+			expect(roomAEvents.some((event) => event.type === 'card.played')).to.equal(true);
+			expect(roomBEvents.some((event) => event.type === 'card.played')).to.equal(false);
```
tweettypography, line 89: **Verify the event type string matches what `announceCardPlayed` actually publishes.**

The existing ring-scoping test (line 52) checks `event.type === 'announce'`, which is what the announcement functions publish into the `RoomEventBus`. If `announceCardPlayed` also publishes `type: 'announce'` (the typical pattern), then `event.type === 'card.played'` here will always be `false` — the room-A "positive" assertion on line 88 will fail, and the room-B "negative" assertion on line 89 will pass vacuously (because the event was never there to begin with, not because the guard filtered it).

If the tests are in fact passing, it means `announceCardPlayed` emits a `'card.played'` typed event to the bus, which is worth a brief comment here to signal the intentional divergence from the `'announce'` convention. If they aren't passing yet (this is a draft), swap to `'announce'` to match the sibling test's pattern and confirm both legs of the assertion are exercising real behavior.

---

packages/engine/src/game.test.ts
```diff
@@ -56,6 +57,42 @@ describe('game.ts', () => {
...
+		} finally {
+			roomA.dispose();
+			roomB.dispose();
+		}
+	});
```
tweettypography, line 159: **Test coverage gap: `creature.levelUp` bleed and the `creature.heal` special path are not exercised.**

This test only proves the `card.played` path is guarded. A second test case that fires `globalSemaphore.emit('creature.levelUp', ...)` with a monster owned by room A would immediately surface the missing `wrapGameEvent` on the `creature.levelUp` listener (see the comment on `announcements/index.ts` lines 141–145). Similarly, `creature.heal` has its guard added inline rather than via the shared `wrapGameEvent` path, so a dedicated assertion there would give confidence that the special-case path stays correct if the heal handler is ever refactored.

Suggested additions alongside this test:

```ts
it('does not mirror level-up announcements across active rooms', async () => {
    // fire globalSemaphore.emit('creature.levelUp', monsterName, monster, { monster, level: 2 })
    // where monster is owned by room A — expect room B to stay silent
});

it('does not mirror heal announcements across active rooms', async () => {
    // fire globalSemaphore.emit('creature.heal', ...) with a room-A-owned creature
});
```

---

packages/engine/src/announcements/index.ts
```diff
@@ -49,53 +49,118 @@ import { announceXPGain } from './xpGain.js';
 import { announceLevelUp } from './level-up.js';
 import type { RoomEventBus } from '../events/index.js';
 
+function createRoomScopedEventGuard(game: any): (...args: any[]) => boolean {
```
tweettypography: **`game: any` weakens the entire guard.**

The parameter type leaks `any` throughout `ownsDirectly` (no IDE or compiler help catching property name typos like `game.characers`). Since `Game` is already typed in `game.ts`, importing and using it here would close the loop — especially important for a security-boundary function like this. Even a narrow structural type (e.g. `{ ring: object; exploration: object; characters: Record<string, { monsters: unknown[] }> }`) would catch the obvious mistakes without a full circular import.

---

tweettypography (commented):
Good diagnosis — the global semaphore is exactly the right place to intercept cross-room bleed, and gating by entity ownership is the correct mechanism. The ring-event rename (`wrap` → `wrapRingEvent` vs `wrapGameEvent`) is a nice clarity improvement, and the `creature.heal` guard is properly added inline. The new test structure (try/finally with dispose, direct `globalSemaphore.emit`) is solid.

One functional bug needs addressing before this is safe to merge, plus a few secondary concerns:

---

**Must fix**

- **`creature.levelUp` is unguarded.** It's the only entry in `gameEvents` that doesn't use `wrapGameEvent`. A level-up event from room A's creature will still fire `announceLevelUp` into room B's `eb`. See inline comment on lines 141–145.

---

**Should fix**

- **Test positive-assertion may be vacuous.** The adjacent ring-scoping test asserts `event.type === 'announce'`; if `announceCardPlayed` follows the same convention the room-A `expect(...).to.equal(true)` will fail (or the test passes for the wrong reasons). See inline comment on lines 86–89.

- **`creature.levelUp` and `creature.heal` paths not covered by tests.** The unguarded levelUp handler would be caught immediately if there were a test for it. See inline comment on the test block.

---

**Worth addressing (not blockers for this fix)**

- **`ownsDirectly` is O(players × monsters × collections) per event** — fires on every `card.rolled`, `creature.hit`, etc. A `Set<object>` rebuilt on `stateChange` would make it O(1). See inline comment on lines 53–80.
- **`WeakSet` + `walk` closure allocated per invocation** — avoidable with a fast top-level args pass before falling back to the deep walk. Magic number `3` should be a named constant. See inline comment on lines 82–87.
- **`game: any` parameter** — the guard is a correctness boundary; a structural type here catches property typos at compile time. See inline comment on line 52.

---

PR #333, opened 2026-04-14
Implement card management engine and server foundations

tweettypography:
## Summary
- Implemented card-management foundations in the engine: unified card/inventory views, unequip/move flows, preset save/load/delete APIs, new command handlers, and command catalog updates.
- Added server tRPC procedures for workshop operations: `game.myInventory`, `game.unequipCard`, `game.unequipAll`, `game.equipCards`, `game.moveCard`, `game.savePreset`, `game.loadPreset`, and `game.deletePreset`.
- Added targeted engine and server tests for the new behaviors.

## Testing
- `pnpm --filter @deck-monsters/engine test`
- `pnpm --filter @deck-monsters/server test`

## Notes
- Web workshop UI route/components/hooks are in progress in follow-up commits on this same branch.

<div><a href="https://cursor.com/agents/bc-96f3cdf5-62a3-487b-b2cd-98688599721d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-96f3cdf5-62a3-487b-b2cd-98688599721d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/apps/web/src/__tests__/monsterWorkshopPanel.dragdrop.test.tsx
+++ b/apps/web/src/__tests__/monsterWorkshopPanel.dragdrop.test.tsx
@@ -1,0 +1,85 @@
+import { fireEvent, render, screen } from '@testing-library/react';
+import { describe, expect, it, vi } from 'vitest';
+import MonsterWorkshopPanel from '../components/MonsterWorkshopPanel.js';
+
+const baseMonster = {
+  name: 'Stonefang',
+  type: 'Basilisk',
+  level: 3,
+  inRing: true,
+  inEncounter: false,
+  cardSlots: 1,
+  cards: [null] as Array<string | null>,
+  presets: {},
+};
+
+describe('MonsterWorkshopPanel drag/drop lock behavior', () => {
+  it('allows dropping into monster slots while in ring but not in encounter', async () => {
+    const onDropCard = vi.fn(async () => undefined);
+    render(
+      <MonsterWorkshopPanel
+        monster={{
+          ...baseMonster,
+          cards: [],
+        }}
+        selectedMonsterName={null}
+        selectedCard={null}
+        onDropCard={onDropCard}
+        onTapSlot={() => undefined}
+        onSelectCard={() => undefined}
+        onSavePreset={() => undefined}
+        onLoadPreset={() => undefined}
+        onDeletePreset={() => undefined}
+      />,
+    );
+
+    const targetSlot = screen.getByRole('button', { name: 'Empty slot' });
+    expect(targetSlot).not.toHaveAttribute('disabled');
+
+    const dataTransfer = {
+      store: {} as Record<string, string>,
+      setData(type: string, value: string) {
+        this.store[type] = value;
+      },
+      getData(type: string) {
+        return this.store[type] ?? '';
+      },
+      effectAllowed: 'none',
+    };
+    dataTransfer.setData(
+      'application/x-deck-monsters-card',
+      JSON.stringify({ location: { kind: 'inventory' }, cardName: 'Hit' }),
+    );
+
+    fireEvent.dragOver(targetSlot, { dataTransfer });
+    fireEvent.drop(targetSlot, { dataTransfer });
+
+    expect(onDropCard).toHaveBeenCalledTimes(1);
+    expect(onDropCard).toHaveBeenCalledWith({ kind: 'inventory' }, 'Hit');
+  });
+
+  it('keeps monster slots locked during an encounter', () => {
+    const onDropCard = vi.fn(async () => undefined);
+    render(
+      <MonsterWorkshopPanel
+        monster={{
+          ...baseMonster,
+          inRing: true,
+          inEncounter: true,
+          cards: [],
+        }}
+        selectedMonsterName={null}
+        selectedCard={null}
+        onDropCard={onDropCard}
+        onTapSlot={() => undefined}
+        onSelectCard={() => undefined}
+        onSavePreset={() => undefined}
+        onLoadPreset={() => undefined}
+        onDeletePreset={() => undefined}
+      />,
+    );
+
+    const targetSlot = screen.getByRole('button', { name: 'Empty slot' });
+    expect(targetSlot).toHaveAttribute('disabled');
+  });
+});
--- a/apps/web/src/__tests__/monsterWorkshopPanel.dragdrop.test.tsx
+++ b/apps/web/src/__tests__/monsterWorkshopPanel.dragdrop.test.tsx
@@ -9,7 +9,7 @@ const baseMonster = {
   inRing: true,
   inEncounter: false,
   cardSlots: 1,
-  cards: [null] as Array<string | null>,
+  cards: [] as string[],
   presets: {},
 };
 
@@ -22,7 +22,7 @@ describe('MonsterWorkshopPanel drag/drop lock behavior', () => {
           ...baseMonster,
           cards: [],
         }}
-        selectedMonsterName={null}
+        showSelectionHint={false}
         selectedCard={null}
         onDropCard={onDropCard}
         onTapSlot={() => undefined}
@@ -68,7 +68,7 @@ describe('MonsterWorkshopPanel drag/drop lock behavior', () => {
           inEncounter: true,
           cards: [],
         }}
-        selectedMonsterName={null}
+        showSelectionHint={false}
         selectedCard={null}
         onDropCard={onDropCard}
         onTapSlot={() => undefined}
@@ -81,5 +81,30 @@ describe('MonsterWorkshopPanel drag/drop lock behavior', () => {
 
     const targetSlot = screen.getByRole('button', { name: 'Empty slot' });
     expect(targetSlot).toHaveAttribute('disabled');
+  });
+
+  it('treats occupied slot taps as destination taps when a card is selected', () => {
+    const onTapSlot = vi.fn();
+    const onSelectCard = vi.fn();
+    render(
+      <MonsterWorkshopPanel
+        monster={{
+          ...baseMonster,
+          cards: ['Hit'],
+        }}
+        showSelectionHint
+        selectedCard={{ location: { kind: 'inventory' }, cardName: 'Heal' }}
+        onDropCard={() => undefined}
+        onTapSlot={onTapSlot}
+        onSelectCard={onSelectCard}
+        onSavePreset={() => undefined}
+        onLoadPreset={() => undefined}
+        onDeletePreset={() => undefined}
+      />,
+    );
+
+    fireEvent.click(screen.getByRole('button', { name: 'Hit' }));
+    expect(onTapSlot).toHaveBeenCalledWith({ kind: 'monster', monsterName: 'Stonefang' });
+    expect(onSelectCard).not.toHaveBeenCalled();
   });
 });
--- a/apps/web/src/__tests__/useDeckWorkshop.test.ts
+++ b/apps/web/src/__tests__/useDeckWorkshop.test.ts
@@ -1,0 +1,7 @@
+import { describe, expect, it } from 'vitest';
+
+describe('useDeckWorkshop cache helpers', () => {
+	it('placeholder sanity test for workshop hook module', () => {
+		expect(true).toBe(true);
+	});
+});
--- a/apps/web/src/__tests__/useDeckWorkshop.test.ts
+++ b/apps/web/src/__tests__/useDeckWorkshop.test.ts
@@ -1,7 +1,102 @@
-import { describe, expect, it } from 'vitest';
+import { renderHook, act } from '@testing-library/react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
 
-describe('useDeckWorkshop cache helpers', () => {
-	it('placeholder sanity test for workshop hook module', () => {
-		expect(true).toBe(true);
-	});
+const mocks = vi.hoisted(() => {
+  const inventoryInvalidate = vi.fn(async () => undefined);
+  const monstersInvalidate = vi.fn(async () => undefined);
+  const roomInfoUseQuery = vi.fn();
+  const myInventoryUseQuery = vi.fn();
+  const defaultMutation = vi.fn((options?: { onSuccess?: () => Promise<void> }) => ({
+    isPending: false,
+    error: null,
+    mutateAsync: vi.fn(async () => {
+      await options?.onSuccess?.();
+      return { equippedCount: 1, requestedCount: 1, skippedCards: [] };
+    }),
+  }));
+
+  return {
+    inventoryInvalidate,
+    monstersInvalidate,
+    roomInfoUseQuery,
+    myInventoryUseQuery,
+    unequipCardUseMutation: vi.fn(defaultMutation),
+    unequipAllUseMutation: vi.fn(defaultMutation),
+    equipCardsUseMutation: vi.fn(defaultMutation),
+    moveCardUseMutation: vi.fn(defaultMutation),
+    savePresetUseMutation: vi.fn(defaultMutation),
+    loadPresetUseMutation: vi.fn(defaultMutation),
+    deletePresetUseMutation: vi.fn(defaultMutation),
+  };
+});
+
+vi.mock('../lib/trpc.js', () => ({
+  trpc: {
+    useUtils: () => ({
+      game: {
+        myInventory: { invalidate: mocks.inventoryInvalidate },
+        myMonsters: { invalidate: mocks.monstersInvalidate },
+      },
+    }),
+    room: {
+      info: {
+        useQuery: mocks.roomInfoUseQuery,
+      },
+    },
+    game: {
+      myInventory: {
+        useQuery: mocks.myInventoryUseQuery,
+      },
+      unequipCard: { useMutation: mocks.unequipCardUseMutation },
+      unequipAll: { useMutation: mocks.unequipAllUseMutation },
+      equipCards: { useMutation: mocks.equipCardsUseMutation },
+      moveCard: { useMutation: mocks.moveCardUseMutation },
+      savePreset: { useMutation: mocks.savePresetUseMutation },
+      loadPreset: { useMutation: mocks.loadPresetUseMutation },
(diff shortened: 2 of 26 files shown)


packages/server/src/trpc/router.ts
```diff
@@ -445,6 +638,421 @@ export function createRouter(roomManager: RoomManager) {
...
+				const channel = createSilentChannel({ eventBus, userId: ctx.userId, commandId });
+				const result = await roomManager.runSerializedEngineWork(input.roomId, async () => {
+					if (input.replaceAll && typeof character.unequipAll === 'function') {
+						await character.unequipAll({
+							channel,
+							monsterName: input.monsterName,
+						});
+					}
+
+					await character.equipMonster({
```
tweettypography: **Bug: `equipCards` procedure calls the old `equipMonster` engine method, not the new `equipCards` method.**

The guard on line 805 checks for `character.equipMonster`, and this call also uses `equipMonster`. This means all the safeguards added in the new `beastmaster.equipCards()` — per-card slot-count check, `MAX_CARD_COPIES_IN_HAND` enforcement, `canHoldCard` validation, and the `skippedCards` return value — are completely bypassed when the web UI drags a card onto a monster slot.

```ts
// Should be:
await character.equipCards({
  channel,
  monsterName: input.monsterName,
  cardNames: input.cardNames,
  replaceAll: input.replaceAll ?? false,
});
```

Also update the guard on line 805 to check `character.equipCards` instead of `character.equipMonster`.

---

packages/server/src/trpc/router.ts
```diff
@@ -445,6 +638,421 @@ export function createRouter(roomManager: RoomManager) {
...
+					type: 'card.equipped' as EventType,
+					scope: 'private',
+					targetUserId: ctx.userId,
+					text: '',
+					payload: {
+						monsterName: result.monsterName,
+						cards: result.cards,
+					},
+				});
+				return { ok: true };
```
tweettypography: **`equipCards` always returns `{ ok: true }`, silently discarding skipped-card info.**

Once the call is switched to `character.equipCards()` (see above), the engine returns `{ equipped, requested, skippedCards, monsterName }`. The client should receive this so it can surface a message like "Equipped 1/2 — Heal was not in your deck" rather than always showing a success toast.

Compare with how `loadPreset` correctly forwards `equippedCount`, `requestedCount`, and `skippedCards` to the client.

---

apps/web/src/components/CardSlot.tsx
```diff
@@ -0,0 +1,98 @@
...
+      return;
+    }
+    await onTapSlot?.(location);
+  }
+
+  return (
+    <button
+      type="button"
+      className={[
+        'card-slot',
```
tweettypography: **CSS class name mismatch — card slot styling will never apply.**

The component uses `card-slot`, `card-slot-icon`, `card-slot-name`, and `card-slot-class`, but `base.css` defines `.workshop-card-slot`, `.workshop-card-icon`, `.workshop-card-name`, and `.workshop-card-class`. None of the sizing, flex layout, font, or colour rules will be applied.

Either rename the classes in the component to match the CSS, or rename the CSS selectors to match the component — pick one convention and use it consistently.

---

packages/engine/src/characters/beastmaster.ts
```diff
@@ -354,6 +364,608 @@ class Beastmaster extends BaseCharacter {
...
+					reason: "you don't appear to have a monster by that name.",
+				}),
+			)
+			.then((monster) => {
+				if (monster.inEncounter) {
+					return Promise.reject(
+						channel({
+							announce: `You cannot unequip cards from ${monster.givenName} while they are fighting!`,
+						}),
+					) as unknown as { removedCount: number; monsterName: string };
```
tweettypography: **`Promise.reject(channel({…}))` rejects with a `Promise` value, not an `Error`.**

`channel()` returns `Promise<unknown>`, so `Promise.reject(channel({…}))` creates a rejected promise whose _rejection reason_ is itself a pending Promise. Any `catch` handler — including the tRPC router — will receive a `Promise` object instead of an `Error`, making error messages impossible to surface to the user. The server will return a generic 500 for "monster is fighting" or "card not found" cases.

This pattern is repeated in `unequipAll`, `moveCard`, `equipCards`, `savePreset`, `loadPreset`, and `deletePreset`. The consistent fix is to throw a proper Error after sending the announcement:

```ts
await channel({ announce: `You cannot unequip cards from ${monster.givenName} while they are fighting!` });
throw new Error(`Cannot unequip from ${monster.givenName}: in encounter`);
```

Or use a shared `GameError` class so the router can forward the message as a user-visible tRPC error instead of a 500.

---

apps/web/src/utils/cards.ts
```diff
@@ -0,0 +1,63 @@
...
+export function getCardClass(name: string): CardClass {
+  const lower = name.toLowerCase();
+  if (HEAL_KEYWORDS.some((keyword) => lower.includes(keyword))) return 'heal';
+  if (MELEE_KEYWORDS.some((keyword) => lower.includes(keyword))) return 'melee';
+  if (MAGIC_KEYWORDS.some((keyword) => lower.includes(keyword))) return 'magic';
+  return 'utility';
+}
+
+export function classifyCard(name: string): CardClass {
+  return getCardClass(name);
```
tweettypography: **`classifyCard` is a one-line alias for `getCardClass` — one of them should be removed.**

Having two exported names for the same function creates confusion about which to use. `CardSlot.tsx` imports `classifyCard` while the rest of this file uses `getCardClass` internally. Pick one name and delete the other.

---

apps/web/src/utils/cards.ts
```diff
@@ -0,0 +1,63 @@
...
+    case 'melee':
+      return '--workshop-card-melee';
+    case 'magic':
+      return '--workshop-card-magic';
+    case 'heal':
+      return '--workshop-card-heal';
+    default:
+      return '--color-border';
+  }
+}
```
tweettypography: **`getCardClassColorVar` is dead code — the CSS variables it references don't exist.**

The function returns strings like `'--workshop-card-melee'` but no such custom properties are declared anywhere in `base.css` or elsewhere in the diff. The function is also never called. Either wire it up with corresponding CSS variables, or remove it.

---

apps/web/src/__tests__/useDeckWorkshop.test.ts
```diff
@@ -0,0 +1,7 @@
+import { describe, expect, it } from 'vitest';
+
+describe('useDeckWorkshop cache helpers', () => {
+	it('placeholder sanity test for workshop hook module', () => {
```
tweettypography: **Placeholder test provides no coverage and shouldn't be committed.**

`expect(true).toBe(true)` verifies nothing. Either test the actual cache-invalidation helpers in `useDeckWorkshop` (e.g. that `invalidateWorkshop` calls `utils.game.myInventory.invalidate` and `utils.game.myMonsters.invalidate`) or delete this file until there's real coverage to add.

---

tweettypography (commented):
## Overview

This PR adds a solid foundation for the Deck Workshop: unified card/inventory views, unequip/move/preset flows in the engine, matching tRPC procedures, and a React UI. The architecture (tRPC → `runSerializedEngineWork` → engine method → silent channel) is clean and consistent. A few issues need fixing before this merges.

---

## Bugs

**1. `equipCards` procedure calls `equipMonster`, not the new `equipCards` engine method (line 819, `router.ts`)**
The entire body of `beastmaster.equipCards()` — slot-count guard, `MAX_CARD_COPIES_IN_HAND` check, `canHoldCard` validation, and `skippedCards` return — is bypassed. The web UI drops a card onto a monster and the old interactive flow runs instead. See inline comment.

**2. CSS class name mismatch in `CardSlot.tsx` (line 70)**
The component emits `card-slot`, `card-slot-icon`, `card-slot-name`, `card-slot-class`. The stylesheet defines `.workshop-card-slot`, `.workshop-card-icon`, `.workshop-card-name`, `.workshop-card-class`. None of the card slot styling applies. Slots will render as unstyled buttons.

**3. `Promise.reject(channel({…}))` rejects with a `Promise`, not an `Error` (line 474, `beastmaster.ts`)**
This pattern appears in every new engine method (`unequipCard`, `unequipAll`, `moveCard`, `equipCards`, `savePreset`, `loadPreset`, `deletePreset`). When the engine rejects — e.g. "monster is fighting" — the tRPC router catches a `Promise` object, not a message string, and returns a 500 with no user-facing detail. A `GameError` or simple `throw new Error(message)` after sending the channel announcement would fix this across all seven methods.

---

## Code quality

**4. `equipCards` router procedure returns `{ ok: true }` even for partial equip (line 860, `router.ts`)**
Once the call is switched to `character.equipCards()`, the engine returns `skippedCards`. The router should forward that to the client the same way `loadPreset` does (`equippedCount`, `requestedCount`, `skippedCards`), so the UI can show "Equipped 1/2 — Heal was not in your deck" instead of an unconditional success toast.

**5. `classifyCard` is a one-line alias for `getCardClass` (line 31, `cards.ts`)**
Two exported names for the same function. Pick one and delete the other.

**6. `getCardClassColorVar` is dead code (line 35, `cards.ts`)**
The CSS variables it references (`--workshop-card-melee` etc.) are not declared anywhere. Either wire it up or remove it.

**7. Placeholder test in `useDeckWorkshop.test.ts`**
`expect(true).toBe(true)` should not be committed; either add real assertions or delete the file.

---

## Minor notes (non-blocking)

- Each mutation in `router.ts` publishes `card.equipped` twice (operation metadata + updated card list). The pattern works but is verbose — a small helper could reduce repetition.
- No confirmation dialog before `unequipAll`. Clearing all cards from a monster is destructive; a confirmation step would prevent accidents, at least on the web UI.
- `cards-utils.test.ts` only tests `abbreviateCardName`. The keyword-based `getCardClass` classification is heuristic and would benefit from a few tests covering edge cases (multi-keyword names, case insensitivity).

---

apps/web/src/components/CardSlot.tsx
```diff
@@ -0,0 +1,98 @@
...
+    event.dataTransfer.effectAllowed = 'move';
+  }
+
+  async function handleClick() {
+    if (disabled) return;
+    if (cardName) {
+      onSelectCard?.(location, cardName);
+      return;
+    }
+    await onTapSlot?.(location);
```
tweettypography, line 65: **Mobile bug: tap-to-move silently fails when the destination slot is occupied.**

`onTapSlot` is only called when `cardName` is `null` (empty slot). If a player selects a card and then taps a *filled* slot on a monster, the handler runs `onSelectCard` instead — switching the selection, not performing the move. On a monster with a full deck this means there is **no way to tap-move a card there at all**.

The fix is to check whether there is already a selection active and, if so, treat any slot click as a destination tap rather than a new selection:

```tsx
interface CardSlotProps {
  // …
  hasActiveSelection?: boolean;   // add this prop
}

async function handleClick() {
  if (disabled) return;
  if (cardName && !hasActiveSelection) {
    onSelectCard?.(location, cardName);
    return;
  }
  await onTapSlot?.(location);
}
```

Then in `WorkshopView` pass `hasActiveSelection={!!selected}` down through `MonsterWorkshopPanel` and `InventoryPanel` to each `CardSlot`.

---

apps/web/src/components/MonsterWorkshopPanel.tsx
```diff
@@ -0,0 +1,107 @@
...
+        onSave={onSavePreset}
+        onDelete={onDeletePreset}
+        disabled={locked}
+      />
+
+      {selectedMonsterName === monster.name && (
+        <div className="workshop-mobile-hint">
+          Tap destination slot or inventory to move selected card.
+        </div>
+      )}
```
tweettypography, line 106: **Mobile UX: the "tap destination" hint is invisible for the inventory→monster flow.**

`selectedMonsterName` is `null` whenever a card is selected from the inventory, so this hint never appears in the most common mobile scenario (pick a card from your deck, drop it onto a monster). A player has no visual cue that they should tap a monster slot next.

Consider either:
- Rendering the hint whenever `selectedCard !== null` (not just when it's from this specific monster), or
- Moving a global selection indicator into the `WorkshopView` header (e.g. "**Hit** selected — tap a monster slot to equip, or tap inventory to return it.").

---

packages/server/src/trpc/router.ts
```diff
@@ -445,6 +638,421 @@ export function createRouter(roomManager: RoomManager) {
...
+				const commandId = randomUUID();
+				const channel = createSilentChannel({ eventBus, userId: ctx.userId, commandId });
+				const result = await roomManager.runSerializedEngineWork(input.roomId, () =>
+					character.unequipCard({
+						channel,
+						cardName: input.cardName,
+						monsterName: input.monsterName,
+						count: input.count,
+					}),
+				) as { removedCount: number; monsterName: string };
```
tweettypography, line 708: **Engine errors are thrown as proper `Error` instances now (good), but the router still returns a generic 500 to the client.**

When `announceAndThrow` fires (e.g. "monster is fighting", "card not found"), the thrown `Error` propagates out of `runSerializedEngineWork` and tRPC catches it as an unhandled exception — returning `INTERNAL_SERVER_ERROR` with no message. The web UI's `catch` block then shows "Action failed" instead of the actual engine message.

Wrap the `runSerializedEngineWork` call in a try/catch and re-throw as a `TRPCError`:

```ts
let result: { removedCount: number; monsterName: string };
try {
  result = await roomManager.runSerializedEngineWork(input.roomId, () =>
    character.unequipCard({ channel, ... }),
  ) as { removedCount: number; monsterName: string };
} catch (err) {
  throw new TRPCError({
    code: 'BAD_REQUEST',
    message: err instanceof Error ? err.message : 'Operation failed',
  });
}
```

The same pattern applies to `unequipAll`, `moveCard`, `equipCards`, `savePreset`, `loadPreset`, and `deletePreset`.

---

apps/web/src/components/PresetControl.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const presetNames = Object.keys(presets).sort((a, b) => a.localeCompare(b));
+
+  return (
+    <div className="workshop-presets">
+      <label className="workshop-subheading" htmlFor="preset-select">
+        Presets
+      </label>
+      <div className="workshop-preset-row">
+        <select
+          id="preset-select"
```
tweettypography: **Duplicate `id="preset-select"` — one per monster panel on the page.**

When a player has multiple monsters, each `PresetControl` renders a `<select id="preset-select">` and a `<label htmlFor="preset-select">`. Duplicate IDs break label association (only the first element is targeted), cause accessibility problems, and are invalid HTML.

Use a prop-based or generated id instead:

```tsx
// PresetControl receives a stable identifier from its parent
interface PresetControlProps {
  monsterId: string; // e.g. monster.name, slugified
  // …
}

// then in the JSX:
const selectId = `preset-select-${monsterId}`;
<label htmlFor={selectId}>Presets</label>
<select id={selectId} …>
```

---

apps/web/src/__tests__/monsterWorkshopPanel.dragdrop.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+import MonsterWorkshopPanel from '../components/MonsterWorkshopPanel.js';
+
+const baseMonster = {
+  name: 'Stonefang',
+  type: 'Basilisk',
+  level: 3,
+  inRing: true,
+  inEncounter: false,
+  cardSlots: 1,
+  cards: [null] as Array<string | null>,
```
tweettypography: **Type mismatch in test fixture — `cards: [null]` doesn't match `monster.cards: string[]`.**

`MonsterPanelProps` types `monster.cards` as `string[]`, but the `baseMonster` fixture uses `cards: [null] as Array<string | null>`. This will cause a TypeScript error if strict checks are run against test files. The cast is papering over the mismatch rather than fixing it.

The fixture should use an empty array `cards: []` (which matches `string[]`) since the test immediately overrides it anyway. The `[null]` value was likely left over from exploring slot rendering.

---

tweettypography (commented):
## Follow-up review — all previous issues addressed ✓

All five items from the first review are fixed:
- `equipCards` now calls `character.equipCards` with the full slot/copy guards
- `equipCards` returns `{ equippedCount, requestedCount, skippedCards }`
- CSS class names are consistent (`workshop-card-slot` etc.)
- `Promise.reject(channel({…}))` replaced with the clean `announceAndThrow` helper
- `classifyCard` alias and dead `getCardClassColorVar` removed
- `useDeckWorkshop.test.ts` has real cache-invalidation coverage
- New drag/drop test and expanded `getCardClass` tests are a solid addition

Four items remain, with two being mobile-blocking:

---

### Mobile-blocking

**1. Tap-to-move fails silently when the target slot is occupied (line 63, `CardSlot.tsx`)**
`onTapSlot` is only called for empty slots. If a player selects a card and taps a *full* monster slot, the handler runs `onSelectCard` instead — switching the selection rather than performing the move. A monster with a full deck becomes unreachable via touch. See inline comment for a `hasActiveSelection` prop fix.

**2. Mobile selection hint is absent for the inventory→monster flow (`MonsterWorkshopPanel.tsx`, line 104)**
`selectedMonsterName` is `null` when a card is selected from inventory, so the hint never renders for the most common use case. See inline comment.

---

### Non-blocking

**3. Engine error messages don't reach the client (`router.ts`)**
`announceAndThrow` now throws a proper `Error` — good — but the router doesn't catch it and re-throw as a `TRPCError`. All six mutations return a generic 500 for any engine-side rejection. See inline comment for the try/catch pattern.

**4. Duplicate `id="preset-select"` in `PresetControl.tsx` (line 30)**
One `<select id="preset-select">` is rendered per monster panel, producing duplicate IDs and broken label association. Needs a per-monster id. See inline comment.

---

PR #335, opened 2026-04-15
Refine workshop multi-select source switching

tweettypography:


apps/web/src/views/WorkshopView.tsx
```diff
@@ -52,9 +53,36 @@ export default function WorkshopView() {
...
+    const grouped = selectedCards.reduce<Record<string, number>>((all, selection) => {
+      all[selection.cardName] = (all[selection.cardName] ?? 0) + 1;
+      return all;
+    }, {});
+    return Object.entries(grouped)
+      .map(([cardName, count]) => (count > 1 ? `${cardName} x${count}` : cardName))
+      .join(', ');
+  }, [selectedCards]);
+
+  function groupSelectionByCardName(selection: SelectionState[]): Array<{ cardName: string; count: number }> {
```
tweettypography: `groupSelectionByCardName` has no closure dependencies — it only uses its argument. Move it to module scope (above the component) so it isn't re-created on every render and can be imported by tests.

---

apps/web/src/views/WorkshopView.tsx
```diff
@@ -105,15 +133,112 @@ export default function WorkshopView() {
...
+          toMonsterName: target.monsterName,
+          count,
+        });
+        movedCount += result.movedCount;
+      }
+      setMessage(`Moved ${movedCount} cards to ${target.monsterName}.`);
+    }
+  }
+
+  function isSameSource(a: WorkshopCardLocation, b: WorkshopCardLocation): boolean {
```
tweettypography, line 164: `isSameSource` is a pure function with no closure dependencies. Move it to module scope alongside `groupSelectionByCardName`. That also makes it exportable so `workshop-selection-state.test.ts` can import it rather than duplicating it.

---

apps/web/src/views/WorkshopView.tsx
```diff
@@ -105,15 +133,112 @@ export default function WorkshopView() {
...
+
   async function handleSlotClick(target: WorkshopCardLocation) {
-    if (!selected || !roomId) return;
-    const payload = selected;
-    setSelected(null);
-    await handleDrop(payload.location, target, payload.cardName);
+    if (selectedCards.length < 1 || !roomId) return;
+    const firstSource = selectedCards[0]?.location;
+    if (firstSource && isSameSource(firstSource, target)) {
+      setSelectedCards([]);
```
tweettypography: This silently wipes the entire multi-selection when the user taps within the same source area, with no feedback. A user who taps an already-selected card expecting to toggle just that card off will find their whole selection gone.

`handleSelect` already handles per-card toggle-off via `selectionId` matching. Consider falling through to `onSelectCard` here (which calls `handleSelect`) rather than clearing the entire array, so same-source slot taps behave consistently with same-source card taps.

---

apps/web/src/__tests__/workshop-selection-state.test.ts
```diff
@@ -0,0 +1,130 @@
+import { describe, expect, it } from 'vitest';
+import type { WorkshopCardLocation } from '../components/CardSlot.js';
+
+type SelectionState = {
+  location: WorkshopCardLocation;
+  cardName: string;
+  selectionId: string;
+};
+
+function isSameSource(a: WorkshopCardLocation, b: WorkshopCardLocation): boolean {
```
tweettypography: `isSameSource` and `toggleSelection` are copy-pasted here instead of imported from the production source. If `WorkshopView.handleSelect` or `isSameSource` changes, these tests won't catch it — they only validate the local copies.

Export `isSameSource` and the selection-toggle logic from a shared module (or directly from `WorkshopView` if it moves to module scope), then import them here.

---

tweettypography (commented):
Good improvements here. Removing `hasActiveSelection`/`preferSelection` from `CardSlot` and routing all source-switching logic through `WorkshopView.handleSelect` is the right architectural call — the card slot no longer needs to know about selection context, and the global banner now correctly shows for all selection states. The unequip-all button and multi-select batch move are both solid additions.

A few issues remain.

---

**Bug: drag-and-drop leaves stale selection**

`handleDrop` (the drag path at lines 87–134) never calls `setSelectedCards([])`. If a user has cards tap-selected and then drag-drops a card, the selection badge persists in the global banner. Add a clear at the top of `handleDrop` (after the same-source guard):

```tsx
async function handleDrop(source, target, cardName) {
  if (!roomId) return;
  if (source.kind === 'monster' && target.kind === 'monster' && source.monsterName === target.monsterName) return;
  setSelectedCards([]);  // clear any tap-selection when a drag completes
  // ...
}
```

---

**UX: `showSelectionHint` only shown for source monster panel**

`showSelectionHint={selectedCards.length > 0 && selectedMonsterName === monster.name}` is `false` for every monster panel when the selection source is the inventory (because `selectedMonsterName` is `null`). The global banner shows, but no per-panel hint guides the user toward valid drop targets. Consider `showSelectionHint={selectedCards.length > 0}` to show the hint on all monster panels whenever cards are selected.

---

**Ongoing: engine errors still surface as 500**

The router still doesn't wrap `runSerializedEngineWork` calls in try/catch that re-throws as `TRPCError({ code: 'BAD_REQUEST', message: err.message })`. User-visible engine errors ("monster is fighting", "deck is full") reach the client as generic 500s. This was flagged in the PR #333 review and is still unaddressed.

---

tweettypography (commented):
The third commit (`5e12b47`) addressed both bugs I flagged:

**Fixed ✓ — `handleDrop` now clears selection**

`setSelectedCards([])` is called at the top of `handleDrop` (after the same-source guard), covered by the new `clears selected banner after drag/drop actions` regression test. Clean fix.

**Fixed ✓ — `showSelectionHint` now shows on all monster panels**

`showSelectionHint={selectedCards.length > 0}` removes the `selectedMonsterName` condition that was blocking the hint when the selection source was the inventory. The `selectedMonsterName` memo was also cleaned up since it's no longer needed. The `shows monster panel hints for inventory selections` regression test covers this path.

---

Three items from my inline comments remain open:

- **`isSameSource` and `groupSelectionByCardName` inside the component** — these are pure functions with no closure dependencies and are re-created on every render. Both should sit at module scope. (Inline comments already posted.)
- **`workshop-selection-state.test.ts` copies logic instead of importing it** — `isSameSource` and `toggleSelection` are duplicated rather than exported from source, so the tests don't guard against production drift. (Inline comment already posted.)
- **Same-source slot tap silently clears the full selection** — tapping within the same source area (e.g., an empty slot in your own monster's panel) wipes the entire multi-selection with no feedback and no way to deselect just one card. (Inline comment already posted.)

And the ongoing issue — engine errors surfacing as 500s rather than `TRPCError(BAD_REQUEST)` — is still unaddressed in the router.

The core functionality is solid at this point. The items above are quality improvements rather than correctness bugs.

---

PR #337, opened 2026-04-15
Improve workshop card compatibility visibility and filtering

tweettypography:
## Summary
- add same-monster card reordering in Deck Workshop by drag-and-drop, backed by new engine/server `reorderCards` flow and client mutation wiring
- improve workshop usability by making card names consistently visible on desktop cards and preserving monster-filter compatibility visuals
- add explicit return-to-terminal navigation in room-scoped screens: a `Terminal` menu item plus clickable room-name breadcrumb back to `/room/:roomId`
- make ring/console feed auto-scroll less greedy: new events no longer force-scroll when the user is reading older messages; `Latest` (or returning to bottom) re-enables follow behavior
- slow default fight pacing by doubling baseline delay midpoints while keeping existing env-var overrides intact
- address review blockers by deduplicating `card.equipped` publish in `reorderCards` and moving ConsolePane scroll side-effects out of state updaters
- expand tests across engine, server, and web (including feed-scroll behavior checks and reorder publish regression assertions)

## Walkthrough artifacts
<a href="https://cursor.com/agents/bc-4d94c405-473c-4187-9a7e-a507a3953b49/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Fworkshop_reorder_filter_and_card_names_demo.mp4">workshop_reorder_filter_and_card_names_demo.mp4</a>
Workshop: desktop card names + compatibility filter + same-monster reorder.

<a href="https://cursor.com/agents/bc-4d94c405-473c-4187-9a7e-a507a3953b49/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Froom_screen_navigation_to_terminal_demo.mp4">room_screen_navigation_to_terminal_demo.mp4</a>
Navigation: Workshop/Fight Log back to main Terminal via new menu item and room breadcrumb link.

Desktop card names visible in workshop
Stonefang filter active with incompatible cards greyed
Same-monster reorder visual position change
Filter cleared with full inventory restored
Workshop header showing Terminal return path
Terminal destination after using Terminal nav
Room-name breadcrumb returning to terminal

<div><a href="https://cursor.com/agents/bc-4d94c405-473c-4187-9a7e-a507a3953b49"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-4d94c405-473c-4187-9a7e-a507a3953b49"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/apps/web/src/__tests__/consolePane-scroll-behavior.test.tsx
+++ b/apps/web/src/__tests__/consolePane-scroll-behavior.test.tsx
@@ -1,0 +1,145 @@
+import { fireEvent, render, screen } from '@testing-library/react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import type { ReactNode } from 'react';
+
+const scrollToIndexMock = vi.fn();
+
+const subscriptionHandlers: {
+  onData?: (event: { id: string; data: any }) => void;
+} = {};
+
+vi.mock('react-virtuoso', () => {
+  const React = require('react');
+
+  const Virtuoso = React.forwardRef((props: any, ref: any) => {
+    React.useImperativeHandle(ref, () => ({
+      scrollToIndex: scrollToIndexMock,
+    }));
+
+    return (
+      <div>
+        <div data-testid="event-count">{props.data?.length ?? 0}</div>
+        <button type="button" onClick={() => props.atBottomStateChange?.(false)}>
+          Mark not at bottom
+        </button>
+        <button type="button" onClick={() => props.atBottomStateChange?.(true)}>
+          Mark at bottom
+        </button>
+      </div>
+    );
+  });
+
+  return { Virtuoso };
+});
+
+vi.mock('../lib/auth-context.js', () => ({
+  useAuth: () => ({ user: { id: 'user-1' } }),
+}));
+
+vi.mock('../hooks/useHandshake.js', () => ({
+  useHandshake: () => ({
+    handleHandshakeEvent: () => undefined,
+  }),
+}));
+
+vi.mock('../lib/command-insert-context.js', () => ({
+  useCommandInsert: () => ({
+    registerInsertFn: () => undefined,
+  }),
+}));
+
+vi.mock('../hooks/useCommandAutocomplete.js', () => ({
+  useCommandAutocomplete: () => [],
+}));
+
+vi.mock('./CommandSuggestions.js', () => ({
+  default: () => null,
+}));
+
+vi.mock('./InlineChoices.js', () => ({
+  default: () => null,
+}));
+
+vi.mock('../lib/trpc.js', () => ({
+  trpc: {
+    game: {
+      consoleHistory: {
+        useQuery: () => ({ data: [] }),
+      },
+      pendingPrompt: {
+        useQuery: () => ({ data: null, refetch: vi.fn(async () => ({ data: null })) }),
+      },
+      myMonsters: {
+        useQuery: () => ({ data: [] }),
+      },
+      command: {
+        useMutation: () => ({ mutateAsync: vi.fn(async () => ({ ok: true })) }),
+      },
+      respondToPrompt: {
+        useMutation: () => ({ mutateAsync: vi.fn(async () => ({ ok: true })) }),
+      },
+      cancelPrompt: {
+        useMutation: () => ({ mutateAsync: vi.fn(async () => ({ ok: true })) }),
+      },
+      cancelFlow: {
+        useMutation: () => ({ mutateAsync: vi.fn(async () => ({ ok: true })) }),
+      },
+      ringFeed: {
+        useSubscription: (_input: unknown, handlers: { onData?: (event: { id: string; data: any }) => void }) => {
+          subscriptionHandlers.onData = handlers.onData;
+        },
+      },
+    },
+  },
+}));
+
+vi.mock('../utils/format-event-text.js', () => ({
+  formatEventText: (text: string) => text,
+}));
+
+vi.mock('../utils/console-history-event-map.js', () => ({
+  mapConsoleHistoryEvent: () => null,
+}));
+
+import ConsolePane from '../components/ConsolePane.js';
+
+describe('ConsolePane scroll behavior', () => {
+  beforeEach(() => {
+    scrollToIndexMock.mockReset();
+    subscriptionHandlers.onData = undefined;
+  });
+
+  it('only follows new events when already at bottom', () => {
+    render(<ConsolePane roomId="11111111-1111-1111-1111-111111111111" isActive onEvent={() => undefined} />);
+
+    fireEvent.click(screen.getByRole('button', { name: 'Mark not at bottom' }));
+    subscriptionHandlers.onData?.({
+      id: 'evt-1',
+      data: {
+        id: 'evt-1',
+        type: 'announce',
+        scope: 'private',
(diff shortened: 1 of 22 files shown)


tweettypography (commented):
## Code + UX Review — PR #337

Good feature set overall — the scroll behaviour fix and card-filter are both long-overdue UX wins. Tests are well-structured and cover the new paths. A few things worth addressing before merge, ranging from a genuine bug to design inconsistencies.

---

### 🚨 Bug — double `card.equipped` event publish in `reorderCards`

`packages/server/src/trpc/router.ts` — the `reorderCards` mutation publishes **two** separate `card.equipped` events to the event bus back-to-back:

1. `{ operation: 'reorderCards', monsterName, fromIndex, toIndex }`
2. `{ monsterName, cards }`

Any subscriber that triggers on `card.equipped` (e.g. the ring-feed or inventory refetch) will fire twice per reorder. This looks unintentional — the first event is never consumed anywhere in the diff. Either merge the payloads into a single publish, or drop the first one.

---

### 🚨 DOM side-effect inside a React state updater — `ConsolePane`

```tsx
// ConsolePane.tsx — addConsoleEvent
setConsoleEvents(prev => {
  const next = [...prev, ev];
  if (autoScroll.shouldFollowRef.current) {
    requestAnimationFrame(() => {      // ← DOM call inside a state updater
      virtuosoRef.current?.scrollToIndex(…);
    });
  }
  return next;
});
```

React may invoke state updater functions more than once in Strict Mode (double-invocation in development). This schedules multiple `scrollToIndex` calls per event, causing jank. Scroll should be triggered in a `useEffect` that watches `consoleEvents.length`, not inside the updater.

---

### Inconsistent auto-scroll implementations

`RingPane` uses Virtuoso's native `followOutput` callback; `ConsolePane` replaces `followOutput={false}` and manually drives scrolling via `useFeedAutoScroll`. Both approaches work, but maintaining two divergent patterns will be painful. Recommend unifying on one strategy — ideally the `useFeedAutoScroll` hook (since it's already extracted and tested), applied to `RingPane` the same way it's applied to `ConsolePane`.

---

### `useFeedAutoScroll` — `enable` and `resetToBottom` are identical

Both callbacks just set `shouldFollowRef.current = true`. Exporting two names for the same operation creates confusion about when to call which. Collapse to one (`enable` or `resetToBottom`), or document the semantic difference if one is intended.

---

### Type assertion on `reorderCards` result is unsafe

```ts
// router.ts
const result = await runSerializedMutation(…) as {
  monsterName: string; fromIndex: number; toIndex: number; cards: string[];
};
```

`runSerializedMutation` returns `unknown`. The cast silently accepts any shape. Add a Zod parse or an `instanceof`-style guard here — consistent with how other mutations in this file are handled.

---

### Index parsing from selectionId strings is fragile

Several places parse slot indices by splitting on `:` and calling `.pop()`:

```ts
// WorkshopView.handleDrop
const sourceIndex = Number.parseInt(sourceSelectionId?.split(':').pop() ?? '', 10);
```

```ts
// CardSlot.handleClick
const slotFromSelectionId = Number.parseInt(selectionId.split(':').at(-1) ?? '', 10);
```

If the selectionId format ever changes (e.g. to a UUID), these silently produce `NaN` and fail without any error surfaced to the user. A dedicated `slotIndex` prop/field passed alongside `selectionId` would be cleaner and easier to refactor later.

---

### `canMonsterHoldCard` defaults to `true` for missing method

```ts
// router.ts
if (typeof canHoldCard !== 'function') {
  // Test doubles and legacy snapshots may not provide this function.
  return true;
}
```

Falling back to `true` means cards will appear compatible with any monster that lacks a `canHoldCard` method, potentially polluting `cardCompatibility` with incorrect data. The comment itself flags this as a code smell. If `canHoldCard` is part of the contract, enforce it rather than silently assume compatibility.

---

### Inline ternary for `compatibilityHint` in JSX is hard to read

```tsx
compatibilityHint={
  selectedInventoryCardName
    ? (isCardCompatibleWithMonster(selectedInventoryCardName, monster.name) ? 'eligible' : 'ineligible')
    : 'none'
}
```

Extract this to a `useMemo` or a named variable above the JSX — same pattern already used for `selectedInventoryCardName` and `compatibleCardCount`.

---

### UX — Filter toggle affordance is non-obvious

Clicking a monster's **name** to toggle an inventory filter is a hidden interaction. Users reading the card won't know the name is a button unless they hover (desktop) or discover it by accident (mobile). Consider:
- A small funnel/filter icon alongside the name, or
- A dedicated "Filter" chip below the monster name

The `aria-pressed` attribute is correct and helpful for screen readers — just the visual affordance needs to match.

---

### UX — Incompatible cards are fully `disabled`, not just visually dimmed

```tsx
// InventoryPanel
disabled={unavailable}
incompatible={unavailable}
```

Setting `disabled` prevents drag, click, and any interaction with filtered-out cards. A user who wants to equip a card to a different monster while a filter is active can't even inspect it. Consider making incompatible cards _visually_ de-emphasised (`incompatible` / `opacity: 0.35`) but still interactive (no `disabled`), and only block the equip action at the drop handler level with a clear error message.

---

### UX — Fight pacing doubles are a notable game behaviour change

The delay midpoints are doubled (very_short: 1500 → 3000 ms, etc.). This is the right call for live-feed readability, but worth calling out explicitly in release notes or the CLAUDE.md known-issues section — players will notice a slower pace and may report it as a regression without context.

---

### Minor / Nit

- `compatibilityHint` is computed in `WorkshopView` only when `selectedCards.length === 1`. The conditional is buried in `selectedInventoryCardName` — worth a short inline comment explaining the single-selection constraint.
- CSS `white-space: normal` + `-webkit-line-clamp: 2` is the correct fix for card name wrapping, but `width: 100%` + `max-width: 100%` on a flex child that already fills its container is redundant.
- `reorderCards` in `beastmaster.ts` uses `Promise.resolve(channel(…)).then(…)` for the no-op `from === to` path while the rest of the class uses `async/await`. Stylistic inconsistency worth aligning.

---

### Summary

| Area | Status |
|---|---|
| Double event publish (`reorderCards`) | 🔴 Fix before merge |
| DOM side-effect in state updater | 🔴 Fix before merge |
| Auto-scroll implementation divergence | 🟡 Recommend unifying |
| `useFeedAutoScroll` duplicate methods | 🟡 Minor |
| Unsafe type assertion | 🟡 Recommend |
| Index parsing fragility | 🟡 Recommend |
| `canMonsterHoldCard` fallback | 🟡 Recommend |
| Filter toggle discoverability | 🟡 UX improvement |
| Incompatible cards fully disabled | 🟡 UX improvement |
| Fight pacing — document the change | 🟢 Low |
| CSS redundancy / style nits | 🟢 Low |

The core features — reorder, compatibility filter, nav improvements, and scroll behaviour — are solid and well-tested. Resolving the two 🔴 issues and considering the UX affordance gap for the filter toggle would make this ready to merge.


---

PR #339, opened 2026-04-17
Pace ring fight sub-events at the source (subEventDelay)

tweettypography:
## Summary

Adds `subEventDelay()` (~1000ms midpoint, jittered via existing range helpers) so narrative events from a single card are spaced in real time instead of bursting through the connector. Respects `DECK_MONSTERS_SKIP_DELAYS=1` for tests.

## Key changes

- **`creatures/health.ts`**: `hit`, `heal`, `die` are async; await `subEventDelay()` after each relevant `emit`; await `die()` from `hit`/`heal` when applicable.
- **`cards/base.ts`**: After `'played'`, await `subEventDelay()` before effects. `applyEffects` is async and sequentially awaits encounter/ring effect hooks so async immobilize logic completes before play continues.
- **`cards/hit.ts`**: Async `effect` with delays after attack roll, damage roll, and miss.
- **`ring/index.ts`**: Success path between cards uses `subEventDelay()` instead of stacking `veryShortDelay` on top of intra-card pacing; error path still uses `veryShortDelay`.
- **Call sites**: Await/async updates where cards called `hit`/`heal` without returning the promise (berserk, horn-gore, boost, curse, prion-disease, blink, immobilize effect); exploration hazard/death cards.
- **Tests**: `delay-times.test.ts` for skip vs real delay; targeting/basilisk tests use `await`; coil/forked-stick/mesmerize tests await async immobilize effects when simulating `applyEffects` manually.

## Verification

`pnpm build` and `pnpm --filter @deck-monsters/engine test` — **451 passing**.

<div><a href="https://cursor.com/agents/bc-68f638c5-c52d-4cec-8639-c58574795bfd"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-68f638c5-c52d-4cec-8639-c58574795bfd"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/packages/engine/src/cards/base.ts
+++ b/packages/engine/src/cards/base.ts
@@ -1,6 +1,7 @@
 import { BaseItem, type BaseItemOptions, type BaseItemStatic } from '../items/base.js';
 import { mapSeries } from '../helpers/promise.js';
 import { ATTACK_PHASE, DEFENSE_PHASE, GLOBAL_PHASE } from '../constants/phases.js';
+import { subEventDelay } from '../helpers/delay-times.js';
 
 export interface CardOptions extends Omit<BaseItemOptions, 'flavors'> {
 	cardClass?: string[];
@@ -83,45 +84,43 @@ export class BaseCard<TOptions extends CardOptions = CardOptions> extends BaseIt
 		return !!(this.cardClass && this.cardClass.includes(cardClass));
 	}
 
-	applyEffects(
+	async applyEffects(
 		player: any,
 		proposedTarget: any,
 		ring: any,
 		activeContestants: any[]
-	): any {
+	): Promise<any> {
 		let card: any = this.clone();
 
 		if (activeContestants) {
-			card = activeContestants.reduce(
-				(contestantCard: any, contestant: any) =>
-					contestant.monster.encounterEffects.reduce((effectCard: any, effect: any) => {
-						const modifiedCard = effect({
-							activeContestants,
-							card: effectCard,
-							phase:
-								contestant.monster === player ? ATTACK_PHASE : DEFENSE_PHASE,
-							player,
-							ring,
-							proposedTarget,
-						});
-						return modifiedCard || effectCard;
-					}, contestantCard),
-				card
-			);
+			for (const contestant of activeContestants) {
+				for (const effect of contestant.monster.encounterEffects) {
+					const modifiedCard = await effect({
+						activeContestants,
+						card,
+						phase:
+							contestant.monster === player ? ATTACK_PHASE : DEFENSE_PHASE,
+						player,
+						ring,
+						proposedTarget,
+					});
+					card = modifiedCard || card;
+				}
+			}
 		}
 
 		if (ring && ring.encounterEffects) {
-			card = ring.encounterEffects.reduce((currentCard: any, effect: any) => {
-				const modifiedCard = effect({
+			for (const effect of ring.encounterEffects) {
+				const modifiedCard = await effect({
 					activeContestants,
-					card: currentCard,
+					card,
 					phase: GLOBAL_PHASE,
 					player,
 					ring,
 					proposedTarget,
 				});
-				return modifiedCard || currentCard;
-			}, card);
+				card = modifiedCard || card;
+			}
 		}
 
 		this.new = card;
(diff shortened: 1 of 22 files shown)


tweettypography (commented):
Implementation is correct and complete. All six planned changes land cleanly, plus the PR goes further in the right ways. Notes below.

---

## `subEventDelay` helper (`delay-times.ts`)

Follows existing file conventions exactly — `skip()`, `parsePositiveInt`, `getRangeFromMidpoint`, `sampleInRange`. No round-shaping or cap applied, which is correct (sub-event delay should be a fixed jitter, not scale with round number). Added to the `delayTimes` default export and `DELAY_ENV_KEYS` cleanup in the test.

The "waits ~1000ms" test is safe: `beforeEach` deletes `DECK_MONSTERS_SKIP_DELAYS` before each test so the skip guard doesn't interfere. `Math.random = () => 0` pins the sample to `min` (~667ms for a 1000ms midpoint), and the `at.least(650)` / `at.most(1400)` window is generous enough for CI variance.

## `health.ts` — async + paced emits

Await placement is correct in all three functions:

- **`hit`**: delay fires after the AC-absorption narration, after the partial-AC-absorption narration, and after `emit('hit')`. State mutations (`self.hp = newHP`, AC zeroing) all happen synchronously before any await. The `return self.die(assailant)` tail-call works because `hit` is async and `die` now returns `Promise<boolean>`.
- **`heal`**: HP mutation is synchronous; delay fires after `emit('heal')`. The `return self.die(self)` path is correct.
- **`die`**: delay fires after `emit('die')` inside the guard, so repeated-die calls on already-dead creatures don't add spurious delays.

The orphaned Promise at `base.ts:161` (`this.heal(1)` in the regen interval) is fine — HP mutates before the first `await`, so state is consistent even if the Promise is never awaited.

## `cards/base.ts` — `applyEffects` refactor (bonus correctness fix)

The plan described this as "await after `'played'`" but the PR goes further and converts `applyEffects`' inner `reduce` calls to `for...of / await`. This is an important correctness fix beyond the plan: the old `reduce` returned the synchronous return value from each encounter-effect call, silently discarding any returned Promise. Cards with async encounter effects — notably `immobilize`'s `ImmobilizeEffect` (which now `await`s `player.hit(2, ...)`) — would have had their async work orphaned under the old code. The sequential `for...of` loop fixes this correctly.

## `cards/hit.ts` — four delay points

Delays after: (1) attack roll, (2) damage roll on success, (3) damage roll on curse-of-loki, (4) `emit('miss')`. In the curse-of-loki path the timing is symmetric with the success path (attack-roll delay → damage-roll delay → hit-emit delay inside `player.hit()`), which feels intentional.

## `ring/index.ts` — inter-card gap

```ts
return new Promise<void>(r => setTimeout(r, 0))
    .then(() => subEventDelay())
    .then(() => next())
```

The `setTimeout(r, 0)` yield before `subEventDelay()` ensures lingering Promise callbacks from the card's async chain settle before the inter-card pause begins. Error path correctly retains `veryShortDelay(round)`.

## Additional cards beyond the plan

The plan flagged berserk, blink, boost, curse, horn-gore, immobilize, and prion-disease as calling `.hit/.heal` outside the normal `effect → mapSeries` path (e.g., inside `effectLoop`, encounter-effect closures, overflow branches). All correctly add `async`/`await`. Without these fixes those calls would have returned a truthy `Promise<boolean>` to callers checking for a plain `boolean` — a subtle live bug.

Cards listed as "no individual update needed" (blast, heal, bad-batch, delayed-hit, cloak-of-invisibility, enchanted-faceswap, fight-or-flight, etc.) all flow through `mapSeries`, which `await`s the mapper result. Returning `target.hit(...)` / `target.heal(...)` without explicit `async`/`await` produces a `Promise<boolean>` that `mapSeries` awaits correctly. No gaps there — verified by reading several of them.

## Exploration cards

`death.ts` and `hazard.ts` both add `async`/`await`. `BaseDiscoveryCard.play` wraps with `Promise.resolve(this.effect(...))` so a returned Promise is correctly awaited even for the default `effect` stub.

## Tests

Targeting and basilisk tests correctly add `await` to `.hit()` / `.die()` calls. The coil/forked-stick/mesmerize tests replace the synchronous `encounterEffects.reduce(...)` pattern with `for...of / await` to match the now-async `ImmobilizeEffect`. All three read cleanly.

---

Overall: clean, correct, well-scoped. The `applyEffects` fix is a welcome bonus. Ready to merge once out of draft.

---

PR #350, opened 2026-04-21
Fix fight summary writes and ring-feed replay after buffer miss

### LINKED ISSUE #342: Bug: Fights not written to fight_summaries — write failures silently swallowed in FightSummaryWriter
## Problem

New fights are not appearing in the fight log. The issue is on the **write side** — `FightSummaryWriter` receives the `ring.fightResolved` event but the DB insert fails silently, permanently losing the fight record.

## Current behaviour

After a fight ends, no new row appears in `fight_summaries`. The fight log page and last-fight ticker show stale data indefinitely. The fight *did* happen (players received XP, monsters levelled up) but there is no record.

## Root causes

All in `packages/server/src/fight-summary-writer.ts`:

### 1. Pending data deleted before the async write completes (~line 69)

```ts
const pending = pendingByRoom.get(roomId)
pendingByRoom.delete(roomId)   // ← deleted here, synchronously
await db.transaction(...)       // ← if this throws, pending is gone forever
```

If the DB transaction fails for any reason, `startedAt` and any pending card-drop data are permanently lost. There is no retry path.

### 2. All errors silently swallowed (~line 44)

```ts
void onFightResolved(db, roomId, event).catch(log)
```

`void` discards the promise. `.catch(log)` logs the error but does not rethrow, retry, or alert. Every DB error — constraint violation, connection failure, type mismatch — is swallowed here.

### 3. Likely UUID type mismatch on `winnerOwnerUserId` / `loserOwnerUserId`

The `fight_summaries` table has:
```sql
winner_owner_user_id uuid,
loser_owner_user_id  uuid,
```

The `ring.fightResolved` payload carries `ownerUserId` from the `Contestant` object in the engine. If this value is a non-UUID string (e.g. a Discord snowflake like `"123456789012345678"` instead of a Supabase profile UUID), Postgres will reject the insert with a type error — which is silently swallowed per point 2.

**Check**: Does `ensureConnectorUser` correctly resolve the Supabase profile UUID before setting `userId` on the Contestant, or does it pass through the external connector ID?

### 4. Concurrent fight race on `pendingByRoom` Map (~line 19)

`pendingByRoom` is a plain `Map<string, { startedAt: Date; cardDropName?: string }>`. If two fights in the same room resolve in close succession (e.g. a 3-monster free-for-all), both `fightResolved` handlers race on the same map entry. The second handler reads the first's `startedAt` (or undefined) and both deletes may collide.

### Step 1 — Don't delete pending data before the write

```ts
// Before
const pending = pendingByRoom.get(roomId)
pendingByRoom.delete(roomId)
await db.transaction(...)

// After
const pending = pendingByRoom.get(roomId)
try {
  await db.transaction(...)
  pendingByRoom.delete(roomId)   // only delete after confirmed write
} catch (err) {
  log.error({ err }, 'fight-summary-writer: DB write failed, pending data preserved for retry')
  throw err
}
```

### Step 2 — Surface errors properly

Replace `void onFightResolved(...).catch(log)` with explicit error handling:

```ts
onFightResolved(db, roomId, event).catch((err) => {
  log.error({ err, roomId }, 'fight-summary-writer: failed to write fight summary')
  // TODO: emit a metric or alert so this doesn't go unnoticed
})
```

### Step 3 — Verify and fix `ownerUserId` type

In the engine's ring code, confirm that `Contestant.userId` is always the Supabase profile UUID (from `profiles.id`), not a connector-specific external ID. If the Discord connector is passing a Discord snowflake as `userId`, `ensureConnectorUser` must resolve it to the profile UUID before the `Contestant` is created.

Checklist:
- [ ] Log the exact Postgres error being swallowed (add a temporary `console.error(err)` in `.catch`)
- [ ] Confirm whether `fight_summaries` rows are missing entirely or present with corrupted data
- [ ] Confirm `Contestant.userId` value format at fight resolution time
- [ ] Fix delete-before-write ordering
- [ ] Fix error surfacing
- [ ] Add a metric/alert for fight-summary write failures so future failures are observable

### Step 4 — Key per-room to prevent concurrent fight races

```ts
// Use a per-room write queue to serialise concurrent fightResolved events
const roomWriteQueues = new Map<string, Promise<void>>()

function enqueueWrite(roomId: string, fn: () => Promise<void>) {
  const prev = roomWriteQueues.get(roomId) ?? Promise.resolve()
  const next = prev.then(fn).catch(log)
  roomWriteQueues.set(roomId, next)
}
```

## Files

- `packages/server/src/fight-summary-writer.ts` — primary fix target
- `packages/server/src/db/schema.ts` — `fight_summaries` schema (check `winnerOwnerUserId` column type)
- `packages/engine/src/ring/index.ts` — `ring.fightResolved` payload construction
- `packages/server/src/auth/connector-users.ts` — `ensureConnectorUser` (verify UUID resolution)
- `supabase/migrations/20260410120000_leaderboard_fight_stats.sql` — table DDL

## Acceptance criteria

- [ ] After a fight ends, a row appears in `fight_summaries` within a few seconds
- [ ] Any write failure produces a visible log line with the full error (not just a single-line `.catch(log)`)
- [ ] No fight is lost due to a delete-before-write race
- [ ] The fight log page shows the completed fight without a manual refresh (that's a separate issue once writes are reliable)

Relates to roadmap `docs/roadmap/10-bug-fixes.md` bug #15.

### LINKED ISSUE #343: Bug: Console pane shows no history on reconnect — ring buffer miss with no DB fallback
## Problem

When a player navigates away from the web app and returns — or loses their connection and reconnects — the console pane is empty. It only shows events that arrive *after* reconnection. Events that occurred while the player was absent are not replayed, even though the server has a replay mechanism.

## Current behaviour

1. Player views the console pane (private event stream).
2. Player closes the tab or loses connection for > a few minutes.
3. Player returns. Console pane is blank.
4. Ring pane may also show a gap in the fight narration.

## Root cause

The reconnect-with-replay path exists but breaks for absences longer than the in-memory ring buffer covers.

### How replay is supposed to work

The web client stores a `lastEventId` cursor. On reconnect, it sends `lastEventId` to the `game.ringFeed` tRPC subscription. The server calls `eventBus.getEventsSince(lastEventId)` and replays events from that point.

### Where it breaks

`getEventsSince` operates on an **in-memory ring buffer** capped at 200 events per room (`RING_BUFFER_SIZE = 200` in `packages/engine/src/events/room-event-bus.ts`).

If the player was away long enough that 200+ events have fired since their `lastEventId`, the cursor is no longer in the buffer. `getEventsSince` does a linear scan, finds nothing, and returns an empty array — **silently, with no indication that history was truncated**.

The server has a `room_events` database table that stores *all* events. There is no fallback to query the database when the ring buffer misses.

### Result

- For short disconnects (< ~200 events): works correctly.
- For longer absences: client receives no replay, console shows nothing.

The client has no way to distinguish "nothing happened" from "events were missed" — both produce an empty replay response.

### Step 1 — DB fallback in `getEventsSince`

When `lastEventId` is not found in the ring buffer, fall back to querying `room_events`:

```ts
// packages/engine/src/events/room-event-bus.ts
getEventsSince(lastEventId: string): { events: GameEvent[]; truncated: boolean } {
  const idx = this.eventLog.findIndex(e => e.id === lastEventId)
  if (idx !== -1) {
    return { events: this.eventLog.slice(idx + 1), truncated: false }
  }
  // Buffer miss — signal truncation; caller should query DB
  return { events: [], truncated: true }
}
```

### Step 2 — Server queries DB on buffer miss

In `packages/server/src/trpc/router.ts`, the `ringFeed` subscription:

```ts
if (input.lastEventId) {
  const { events, truncated } = eventBus.getEventsSince(input.lastEventId)
  if (truncated) {
    // Fall back to DB
    const dbEvents = await db
      .select()
      .from(roomEvents)
      .where(
        and(
          eq(roomEvents.roomId, input.roomId),
          gt(roomEvents.eventId, input.lastEventId)  // eventId is stored in room_events
        )
      )
      .orderBy(asc(roomEvents.id))
      .limit(500)
    // Re-hydrate and yield dbEvents
  } else {
    for (const event of events) yield tracked(event.id, event)
  }
}
```

### Step 3 — Gap event as safety net

If the DB also has no events (retention window passed), yield a synthetic `system.gap` event so the client can show "You were away for a while — some events were missed" rather than silently presenting an empty console.

```ts
yield tracked('gap', {
  id: 'gap',
  type: 'system.gap',
  text: 'Some events were missed while you were away.',
  scope: 'public',
  ...
})
```

### Step 4 — Client displays gap message

In `ConsolePane.tsx`, render `system.gap` events as a styled separator: `── missed events ──`.

## Files

- `packages/engine/src/events/room-event-bus.ts` — `getEventsSince`, `RING_BUFFER_SIZE`
- `packages/server/src/trpc/router.ts` — `ringFeed` subscription, catch-up block
- `packages/server/src/db/schema.ts` — `roomEvents` table (has `eventId` string column alongside auto-increment `id`)
- `apps/web/src/components/ConsolePane.tsx` (or equivalent) — render `system.gap`
- `packages/engine/src/events/types.ts` — add `system.gap` event type

## Acceptance criteria

- [ ] After a short absence (< 200 events), replay works as today
- [ ] After a long absence (> 200 events), the DB fallback provides events up to the retention window
- [ ] If events are beyond the retention window, a visible "missed events" separator appears in the console
- [ ] No silent empty-replay for any absence length

Relates to roadmap `docs/roadmap/10-bug-fixes.md` bug #16.  
See also bug #17 (ring buffer gap detection) which is a sub-problem of this issue.

### LINKED ISSUE #344: Bug: Ring event buffer miss returns empty array silently — no DB fallback, no gap signal
## Problem

When a reconnecting client's `lastEventId` has been evicted from the in-memory ring buffer, `getEventsSince()` returns an empty array with no indication that events were truncated. The server cannot distinguish "nothing happened" from "too many events happened" — so it sends nothing, and the client silently presents an incomplete view.

This is the lower-level infrastructure bug underlying the console-missing-history problem (see related issue). Fixing this enables the DB-fallback path that fully resolves reconnect replay.

## Current behaviour

```ts
// packages/engine/src/events/room-event-bus.ts
getEventsSince(lastEventId: string): GameEvent[] {
  const idx = this.eventLog.findIndex(e => e.id === lastEventId)
  if (idx === -1) return []   // ← silent empty return; caller can't tell why
  return this.eventLog.slice(idx + 1)
}
```

- Buffer holds 200 events (`RING_BUFFER_SIZE = 200`).
- If `lastEventId` was evicted, `findIndex` returns `-1` and the caller gets `[]`.
- Caller (tRPC `ringFeed` subscription) has no way to know whether `[]` means "no new events" or "cursor was evicted."

## Proposed fix

Change the return type to include a `truncated` flag:

```ts
getEventsSince(
  lastEventId: string
): { events: GameEvent[]; truncated: boolean } {
  const idx = this.eventLog.findIndex(e => e.id === lastEventId)
  if (idx === -1) {
    return { events: [], truncated: true }
  }
  return { events: this.eventLog.slice(idx + 1), truncated: false }
}
```

All callers then handle the `truncated` case — either by querying the DB (`room_events` table) or by yielding a `system.gap` sentinel event.

Additionally, expose whether the cursor is *ahead* of the buffer (future ID, indicates client is already up to date) vs *behind* (evicted, indicates missed events). These are two different conditions that currently both map to `idx === -1`.

```ts
getEventsSince(lastEventId: string): EventSinceResult {
  const idx = this.eventLog.findIndex(e => e.id === lastEventId)
  if (idx === -1) {
    // Determine if cursor is ahead or behind
    const newestId = this.eventLog.at(-1)?.id
    const isAhead = lastEventId > (newestId ?? '')  // IDs are monotonically increasing
    return { events: [], truncated: !isAhead, upToDate: isAhead }
  }
  return { events: this.eventLog.slice(idx + 1), truncated: false, upToDate: false }
}
```

## Files

- `packages/engine/src/events/room-event-bus.ts` — `getEventsSince` method (primary change)
- `packages/server/src/trpc/router.ts` — `ringFeed` subscription (update caller to handle `truncated`)
- `packages/engine/src/events/types.ts` — add `EventSinceResult` type if needed

## Notes

- This is a prerequisite for the DB-fallback fix in the console-history issue.
- `RING_BUFFER_SIZE = 200` could also be increased (e.g. to 500–1000) as a cheap short-term mitigation for active rooms, but the DB fallback is the correct long-term fix.
- The `room_events` table already stores all events with an `eventId` string column — the infrastructure for DB fallback is there, just not wired to `getEventsSince`.

## Acceptance criteria

- [ ] `getEventsSince` returns `{ truncated: true }` (not `[]`) when the cursor is not in the buffer
- [ ] All callers updated to handle the new return type
- [ ] No silent empty-replay when the ring buffer has rolled past the client's cursor

Relates to roadmap `docs/roadmap/10-bug-fixes.md` bug #17.  
Prerequisite for: console-missing-history issue (bug #16).

tweettypography:
## Summary

This change addresses three related production issues around fight log persistence and reconnect replay.

### #342 — Fight summaries not written

- **Per-room write queue** so concurrent `ring.fightResolved` handlers cannot corrupt shared `pendingByRoom` state.
- **Delete pending metadata only after** the `fight_summaries` transaction succeeds, so a failed insert can be retried on the next resolution path without losing `startedAt` / card-drop data.
- **Owner user IDs**: only UUID-shaped values are written to `winner_owner_user_id` / `loser_owner_user_id` (avoids Postgres errors when a non-UUID connector id slips through).
- **Errors**: structured `log.error` with optional stack and Postgres `code`, plus **`dm_fight_summary_write_failures_total`** counter (no duplicate `legacyLog` on write failure).

### #343 / #344 — Console / ring reconnect and buffer miss

- **`RoomEventBus.getEventsSince`** now returns `{ events, truncated, upToDate }` (exported as `EventsSinceResult`). Empty in-memory buffer after restart returns **non-truncated** so reconnects do not spuriously hit DB/gap.
- **`game.ringFeed`**: on truncation (and not up-to-date), loads catch-up from **`room_events`** via **`getEventsSinceForRingFeed`** (room + visibility: public or private for that user). The 7d fallback applies **`event_id > lastEventId` in SQL** with `orderBy(asc)` + `limit`. Metrics: **`dm_ring_feed_replay_from_db_total`**, **`dm_ring_feed_replay_gap_total`**.
- If the DB has nothing after the cursor (e.g. retention), yields a private **`system.gap`** event with visible text instead of a silent empty replay.
- **Web**: `ConsolePane` shows `system.gap`; `mapConsoleHistoryEvent` maps it for HTTP history; `shouldRenderRingEvent` hides it from the public ring feed.
- **Event persister**: `system.gap` is ephemeral (not stored in `room_events`).

### Review follow-up (post-merge request feedback)

- Empty `eventLog` edge case fixed.
- 7d branch SQL filter aligned with 24h branch.
- **`getEventsSinceForRingFeed`** unit tests** in `room-manager.test.ts`.
- **`room_events` indexing**: migrations already define `room_events_event_id_idx` and unique `(room_id, event_id)` — noted inline at anchor query.

## Verification

- `pnpm build`
- `pnpm --filter @deck-monsters/engine test`
- `pnpm --filter @deck-monsters/server test`
- `pnpm --filter @deck-monsters/web test`

Closes #342. Closes #343. Closes #344.

<div><a href="https://cursor.com/agents/bc-63306d9f-cb04-48ad-92ae-abcc97e61533"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-63306d9f-cb04-48ad-92ae-abcc97e61533"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/apps/web/src/__tests__/console-history-event-map.test.ts
+++ b/apps/web/src/__tests__/console-history-event-map.test.ts
@@ -31,4 +31,19 @@ describe('mapConsoleHistoryEvent', () => {
       text: '-- reconnecting --',
     });
   });
+
+  it('maps system.gap to a system row', () => {
+    const mapped = mapConsoleHistoryEvent({
+      id: 'gap-1',
+      type: 'system.gap',
+      text: '── missed ──',
+      payload: { reason: 'buffer_or_retention' },
+    });
+
+    expect(mapped).toEqual({
+      id: 'gap-1',
+      type: 'system',
+      text: '── missed ──',
+    });
+  });
 });
--- a/apps/web/src/components/ConsolePane.tsx
+++ b/apps/web/src/components/ConsolePane.tsx
@@ -365,8 +365,7 @@ export default function ConsolePane({ roomId, isActive, onEvent }: ConsolePanePr
         // Only process events targeted to this user
         const isPrivate = event.scope === 'private' && event.targetUserId === user?.id;
         const isPublicSystem = event.scope === 'public' && event.type === 'system';
-        const isGapNotice = event.type === 'system.gap' && isPrivate;
-        if (!isPrivate && !isPublicSystem && !isGapNotice) return;
+        if (!isPrivate && !isPublicSystem) return;
 
         onEvent?.(event);
         if (MONSTER_REFRESH_EVENT_TYPES.has(event.type)) {
--- a/apps/web/src/components/ConsolePane.tsx
+++ b/apps/web/src/components/ConsolePane.tsx
@@ -365,7 +365,8 @@ export default function ConsolePane({ roomId, isActive, onEvent }: ConsolePanePr
         // Only process events targeted to this user
         const isPrivate = event.scope === 'private' && event.targetUserId === user?.id;
         const isPublicSystem = event.scope === 'public' && event.type === 'system';
-        if (!isPrivate && !isPublicSystem) return;
+        const isGapNotice = event.type === 'system.gap' && isPrivate;
+        if (!isPrivate && !isPublicSystem && !isGapNotice) return;
 
         onEvent?.(event);
         if (MONSTER_REFRESH_EVENT_TYPES.has(event.type)) {
@@ -378,6 +379,15 @@ export default function ConsolePane({ roomId, isActive, onEvent }: ConsolePanePr
           addConsoleEvent({
             id: event.id,
             type: 'input',
+            text: event.text,
+          });
+          return;
+        }
+
+        if (event.type === 'system.gap') {
+          addConsoleEvent({
+            id: event.id,
+            type: 'system',
             text: event.text,
           });
           return;
--- a/apps/web/src/utils/console-history-event-map.ts
+++ b/apps/web/src/utils/console-history-event-map.ts
@@ -18,6 +18,9 @@ export function mapConsoleHistoryEvent(event: ConsoleHistoryEvent): ConsoleHisto
   if (event.type === 'system' && payload.consoleInput) {
     return { id: event.id, type: 'input', text: event.text };
   }
+  if (event.type === 'system.gap') {
+    return { id: event.id, type: 'system', text: event.text };
+  }
   if (event.type === 'announce' || event.type === 'system') {
     return { id: event.id, type: event.type as 'announce' | 'system', text: event.text };
   }
--- a/apps/web/src/utils/ring-feed-events.test.ts
+++ b/apps/web/src/utils/ring-feed-events.test.ts
@@ -25,4 +25,8 @@ describe('shouldRenderRingEvent', () => {
     expect(shouldRenderRingEvent(event('ring.fight'))).to.equal(true);
     expect(shouldRenderRingEvent(event('ring.xp'))).to.equal(true);
   });
+
+  it('hides private reconnect gap notice', () => {
+    expect(shouldRenderRingEvent(event('system.gap'))).to.equal(false);
+  });
 });
--- a/apps/web/src/utils/ring-feed-events.ts
+++ b/apps/web/src/utils/ring-feed-events.ts
@@ -3,6 +3,8 @@ import type { GameEvent } from '@deck-monsters/server/types';
 const HIDDEN_RING_EVENT_TYPES = new Set<string>([
   // Internal analytics event; users already see flavorful fight-conclusion text.
   'ring.fightResolved',
+  // Private console reconnect notice; never belongs in the public ring feed.
+  'system.gap',
 ]);
 
 export function shouldRenderRingEvent(event: Pick<GameEvent, 'type'>): boolean {
--- a/packages/engine/src/events/index.ts
+++ b/packages/engine/src/events/index.ts
@@ -1,2 +1,2 @@
 export { RoomEventBus } from './room-event-bus.js';
-export type { GameEvent, EventType, EventScope, EventSubscriber } from './types.js';
+export type { GameEvent, EventType, EventScope, EventSubscriber, EventsSinceResult } from './types.js';
--- a/packages/engine/src/events/room-event-bus.test.ts
+++ b/packages/engine/src/events/room-event-bus.test.ts
@@ -211,4 +211,12 @@ describe('RoomEventBus getEventsSince', () => {
 		expect(r.truncated).to.equal(false);
 		expect(r.upToDate).to.equal(true);
 	});
+
+	it('does not mark truncated when the buffer is empty (fresh bus)', () => {
+		const bus = new RoomEventBus(ROOM_ID);
+		const r = bus.getEventsSince('any-cursor');
+		expect(r.events).to.have.length(0);
+		expect(r.truncated).to.equal(false);
+		expect(r.upToDate).to.equal(false);
+	});
 });
--- a/packages/engine/src/events/room-event-bus.test.ts
+++ b/packages/engine/src/events/room-event-bus.test.ts
@@ -157,3 +157,58 @@ describe('RoomEventBus pending prompt snapshots', () => {
 		expect(bus.getPendingPromptForUser('nobody')).to.equal(null);
 	});
 });
+
+describe('RoomEventBus getEventsSince', () => {
+	it('returns slice after id when cursor is in the buffer', () => {
+		const bus = new RoomEventBus(ROOM_ID);
+		const a = bus.publish({
+			type: 'announce',
+			scope: 'public',
+			text: 'a',
+			payload: {},
+		});
+		const b = bus.publish({
+			type: 'announce',
+			scope: 'public',
+			text: 'b',
+			payload: {},
+		});
+		const c = bus.publish({
+			type: 'announce',
+			scope: 'public',
+			text: 'c',
+			payload: {},
+		});
+		const r = bus.getEventsSince(a.id);
+		expect(r.truncated).to.equal(false);
+		expect(r.upToDate).to.equal(false);
+		expect(r.events.map(e => e.id)).to.deep.equal([b.id, c.id]);
+	});
+
+	it('marks truncated when id was evicted and cursor is behind newest', () => {
+		const bus = new RoomEventBus(ROOM_ID);
+		const oldIds: string[] = [];
+		for (let i = 0; i < 210; i++) {
+			const e = bus.publish({
+				type: 'announce',
+				scope: 'public',
+				text: `m${i}`,
+				payload: {},
+			});
+			if (i === 0) oldIds.push(e.id);
+		}
+		const r = bus.getEventsSince(oldIds[0]!);
+		expect(r.events).to.have.length(0);
+		expect(r.truncated).to.equal(true);
+		expect(r.upToDate).to.equal(false);
+	});
+
+	it('marks upToDate when cursor is lexicographically newer than buffer tail', () => {
+		const bus = new RoomEventBus(ROOM_ID);
+		bus.publish({ type: 'announce', scope: 'public', text: 'x', payload: {} });
+		const r = bus.getEventsSince('9999999999999-zzzzzzzz');
+		expect(r.events).to.have.length(0);
+		expect(r.truncated).to.equal(false);
+		expect(r.upToDate).to.equal(true);
+	});
+});
--- a/packages/engine/src/events/room-event-bus.ts
+++ b/packages/engine/src/events/room-event-bus.ts
@@ -70,6 +70,10 @@ export class RoomEventBus {
 		if (idx !== -1) {
 			return { events: this.eventLog.slice(idx + 1), truncated: false, upToDate: false };
 		}
+		// Fresh room after restart (or no events yet): do not treat as buffer truncation.
+		if (this.eventLog.length === 0) {
+			return { events: [], truncated: false, upToDate: false };
+		}
 		const newestId = this.eventLog.at(-1)?.id;
 		const isAhead = newestId !== undefined && eventId > newestId;
 		return { events: [], truncated: !isAhead, upToDate: isAhead };
--- a/packages/engine/src/events/room-event-bus.ts
+++ b/packages/engine/src/events/room-event-bus.ts
@@ -1,12 +1,7 @@
 import { randomUUID } from 'node:crypto';
 
-import type { GameEvent, EventSubscriber, EventScope, EventType } from './types.js';
+import type { GameEvent, EventSubscriber, EventScope, EventType, EventsSinceResult } from './types.js';
 
-// TODO(ordering): When a client reconnects with a lastEventId that has already
-// been evicted from this buffer, getEventsSince() returns [] with no signal
-// that history was truncated.  The client silently presents an incomplete
-// replay.  Fix: return a synthetic gap event (or throw a typed error) when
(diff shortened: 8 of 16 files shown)


tweettypography (commented):
Good overall direction — the three root-cause fixes (delete-after-commit, per-room write queue, truncation signal) are all correct. A few issues before this is ready to merge.

---

## Issues

### 1. `getEventsSince` returns `truncated: true` on empty buffer (bug)

`packages/engine/src/events/room-event-bus.ts`

```ts
const newestId = this.eventLog.at(-1)?.id;        // undefined when buffer is empty
const isAhead = newestId !== undefined && eventId > newestId;  // always false
return { events: [], truncated: !isAhead, upToDate: isAhead }; // truncated: true ❌
```

If the server restarts and a client reconnects before any events are published, every cursor comes back `truncated: true`. This fires the DB fallback, the DB also returns nothing, and the user sees the gap notice despite not actually missing anything — the server just restarted.

Fix:
```ts
if (this.eventLog.length === 0) {
    return { events: [], truncated: false, upToDate: false };
}
```

### 2. `getEventsSinceForRingFeed` — 7d fallback applies `limit` before filtering (correctness)

`packages/server/src/room-manager.ts`

```ts
rows = await this.db
    .select()
    .from(roomEvents)
    .where(and(eq(roomEvents.roomId, roomId), visibility, gte(roomEvents.createdAt, cutoff7d)))
    .orderBy(desc(roomEvents.id))
    .limit(limit);                               // ← applied before filter

return rows.reverse().map(dbRowToGameEvent).filter(e => e.id > lastEventId).slice(0, limit);
```

`limit` is applied DESC in the DB and then filtered in memory. If many of the newest `limit` rows predate `lastEventId`, the returned slice has fewer than `limit` items even when qualifying rows exist further back in the 7d window. The 24h branch already correctly applies `gt(roomEvents.eventId, lastEventId)` in the WHERE clause — the 7d branch should too:

```ts
.where(and(eq(roomEvents.roomId, roomId), visibility, gte(roomEvents.createdAt, cutoff7d), gt(roomEvents.eventId, lastEventId)))
.orderBy(asc(roomEvents.id))
.limit(limit)
```

### 3. Missing index on `room_events.event_id` (performance risk)

The anchor lookup in `getEventsSinceForRingFeed` queries by `event_id` on every reconnect after a buffer miss. If there's no index (or composite `(room_id, event_id)` index) on that column, this is a full table scan on what can be a large table. Worth verifying or adding one.

### 4. No tests for `getEventsSinceForRingFeed`

This is the most complex new method — three DB query branches, lexicographic ordering assumptions on `event_id`, and visibility filtering. It has zero test coverage. Given that #343/#344 were silent failures (wrong data, no error), the core regression path — buffer miss → DB fallback → correct events returned for the reconnecting user and not others — should have at least a basic integration test.

---

## Minor

- **`isGapNotice` is logically redundant** (`ConsolePane.tsx:368`): `system.gap` events are scoped `private` targeted to a specific user, so `isPrivate` is already `true` for them. `isGapNotice` can only be `true` when `isPrivate` is already `true`, making the third clause in the guard always redundant.

- **Double logging on fight summary write failure** (`fight-summary-writer.ts`): both `log.error(...)` and `legacyLog(err)` fire on write failure. If the caller's `legacyLog` logs to the same destination, each failure produces two entries.

---

## What's well done

- **Delete-after-commit** is the right fix and is clean.
- **Per-room write queue** with the `catch` to keep the chain alive after a prior error is exactly right.
- **`profileUuidOrNull`** correctly fixes the Postgres UUID type error with Discord snowflakes.
- **`EventsSinceResult`** struct is a clean API upgrade — `truncated`/`upToDate` are self-documenting.
- **`system.gap` in `EPHEMERAL_TYPES`** is correct — no replay value.
- **Three new unit tests for `getEventsSince`** cover the important states well.
- **New metrics** (`dm_fight_summary_write_failures_total`, `dm_ring_feed_replay_from_db_total`, `dm_ring_feed_replay_gap_total`) will make these failure modes visible in production.

---

tweettypography (commented):
All six issues from the first pass are addressed. LGTM.

- **Empty buffer** — early-return guard in `getEventsSince` is correct, test added.
- **7d fallback** — `gt(roomEvents.eventId, lastEventId)` now in the SQL WHERE clause with `orderBy(asc)`, in-memory filter removed.
- **Index** — anchor query comment documents the existing `room_events_event_id_idx` / unique `(room_id, event_id)` constraint.
- **`getEventsSinceForRingFeed` tests** — anchor hit, anchor miss (24h), 7d fallback, and private-only visibility all covered.
- **Double logging** — `legacyLog` removed from the `fightResolved` error path; structured `log.error` is the sole emitter there.
- **`isGapNotice` redundancy** — variable gone; gap events pass through `isPrivate` naturally and hit the dedicated handler block.

Ready to merge.

---

Issue #25, opened 2017-05-17
Adventures


cormacmccarthy:
For world exploration, it would be cool to have a "jobs request board" in the Inn in town. Or maybe the characters are all in a guild, and that's where the board is. Anyways, you take a job, get some details about it and the things you are going to encounter/fight, and then you build your adventure and your fighting decks and send your monster out to try and complete the job.

Really for this you probably wouldn't even need the adventure deck yet, it could easily be a slightly modified ring with a series of fights with some flavor text between them with maybe some basic rolls. This could be a great start for our PVE portion of the game.

PR #24, opened 2017-05-16
Adds capability to remove monster from ring, as well as fight start timer

cormacmccarthy:
fixes #19

Issue #23, opened 2017-05-15
Card pool display


cormacmccarthy:
Currently when you see your card list, it looks something like this:

hit
hit
heal
flee
hit
heal
hit
hit
flee
harden
heal

That should instead display like this:
hit (5)
heal (3)
flee (2)
harden

Issue #22, opened 2017-05-15
look at things


cormacmccarthy:
`jane look at monsters` - Should show you all of your monsters
`jane look at cards` - Should show you all of your cards

PR #355, opened 2026-07-31
fix(engine,server): resolve bug-doc #15–#18, #20–#25 — timing, sync, reconnect replay, and cross-room correctness

tweettypography:
Resolves bug-doc items **#15, #16, #17, #18, #20, #21, #22, #23, #24, #25** (see `docs/roadmap/10-bug-fixes.md` for full write-ups of each), plus all four points from the requested-changes review. Records **#26** (process-wide shop singleton) for a design decision without fixing it.

## Engine

- **#20 — fight pacing**: `Ring.fight()`'s normal card-play path paced with `subEventDelay` (~1s) instead of the configured `veryShortDelay` (2–4s) every other fight path uses, so live fights scrolled by too fast to follow. `DECK_MONSTERS_SKIP_DELAYS` behavior unchanged.
- **#20 — multi-step command crashes**: cancelling a flow resolved prompts with the literal `'__cancelled__'` string that game code parsed as a selection, and `chooseItems` only accepted numeric indices (a typed card name became `NaN` and aborted the whole flow). The engine now exports `PROMPT_CANCELLED`/`PromptCancelledError`; `chooseItems` accepts indices or case-insensitive names and skips invalid entries instead of aborting.
- **#25 — CRITICAL, cross-room reward duplication**: `Game`'s `creature.win`/`loss`/`permaDeath`/`fled` listeners sat unscoped on the process-wide semaphore, so one fight's outcome granted XP, coins, and a drawn card **once per currently-loaded room**. Now wrapped with the room-scoping guard. Also: `Ring.handlePermaDeath` was missing the emit its siblings have, so perma-death owners previously got *no* reward at all.
- **#22/#23/#24**: boss despawn timers are now tracked and disposed; `stateChange` save scheduling is room-scoped (one room's activity can no longer reset another's save debounce); direct in-place mutations in `getCharacter`, `Ring.removeMonster`, and `removeItem` now go through setters so they persist across restarts. The room-scoping guard itself was hardened against a reentrancy stack-overflow (it now reads raw `optionsStore`, never live lazy-initializing getters).

## Server

- **#16/#17 — reconnect replay**: a cold ring buffer (the state after every restart/idle eviction) reported `truncated: false`, suppressing the durable-storage fallback entirely — returning players silently got nothing. `EventsSinceResult` now carries a `found/ahead/evicted/cold` status; cold replays hit the DB without false gap warnings, evicted-and-empty warns. Synthetic frame ids (`handshake`/`heartbeat`/gap) now use the standard `${epochMs}-` shape so a stray cursor stays resolvable.
- **#15 — fight summaries**: the pending-fight snapshot is captured synchronously at `fightResolved` delivery (fixing the cross-fight clobber race), writes retry with backoff instead of dropping rows on transient DB errors, and per-room maps are cleaned on detach.
- **#18 — quick actions**: new `buildQuickActions` emits up to four contextual command chips after each command settles, using real `COMMAND_CATALOG` syntax; excluded from event persistence.
- **#20 — command starvation**: interactive command actions moved from the room-wide serialized lane to per-`roomId:userId` lanes; workshop mutations fail fast with `PRECONDITION_FAILED` while the caller has an active console flow instead of hanging behind it.
- **#21 — idle sweep**: `unloadRoom` no longer unloads a room mid-fight (retries next sweep), so in-progress fights can't lose their announcements/stats/summary.

## Review fixes (requested changes)

1. **Replay race**: the ringFeed live subscriber now attaches *before* the replay is assembled (buffered duplicates dropped by id on drain; try/finally opens before the replay so a throw can't leak the subscriber), and `getEventsSinceForRingFeed` paginates past the old single-query 500-row cliff, surfacing `limitReached` so a `system.gap` marker is emitted instead of a silently truncated stream.
2. **Retry-after-reset**: detaching the fight-summary writer now cancels queued/backoff retries, so no write can land after `resetRoomState()` clears summaries and resets the counter.
3. **Flow-lock race**: `activeFlows` is now a Map keyed by flow token; a cancelled flow's late-settling cleanup only clears a lock it still owns, so it can no longer release a newer command's lock. `cancelFlow`'s force-clear stays unconditional.
4. **Doc drift**: `CLAUDE.md` and `AGENTS.md` no longer list #15/#16 as active; both point at the actual root causes and the still-open #26.

## Testing

680 tests pass across all packages (480 engine / 105 server / 71 web / 20 discord / 4 harness); build and lint clean. New coverage includes: per-user command lanes, workshop fail-fast, the cancellation-overlap scenario, quick-action emission, subscribe-before-replay ordering + dedup, replay pagination and the replay-cap gap marker, retry cancellation on detach, named/cancelled/invalid item-selection parsing, cross-room reward isolation, and a full fight through the real non-skipped pacing path (previously zero coverage on the non-skip branches).

Architecture notes for future maintainers are in `docs/engine-concurrency-and-timing.md`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01MnewFgP4nZXktp3iqj9NVJ

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -128,7 +128,6 @@ Missing a room filter is a recurring source of bugs. See [`docs/room-scoping.md`
 
 ### Active known bugs
 
-Two open bugs in the real-time data layer — do not close or work around without fixing the root cause:
+- **Card shop is a process-wide singleton** — every room shares one shop inventory and closing time (`packages/engine/src/items/store/shop.ts`), violating the room-scoping rule above. Needs a design decision (per-room shop vs. intentionally shared) before implementation. See bug #26 in `docs/roadmap/10-bug-fixes.md`.
 
-- **Fight log stale after new fights** — the fight log page (`/room/:roomId/fights`) doesn't update when new fights complete. Subscription or query-cache invalidation is broken. See bug #15 in `docs/roadmap/10-bug-fixes.md`.
-- **Console missing history on reconnect** — the console pane doesn't replay events from while the user was away. The reconnect-with-replay path exists but isn't delivering historical data. See bug #16 in `docs/roadmap/10-bug-fixes.md`.
+Previously listed here and now fixed: fight log stale after new fights (#15 — the root cause was write-side races and dropped retries in `fight-summary-writer.ts`, not query-cache invalidation) and console history missing on reconnect (#16/#17 — the engine's cold ring buffer reported "nothing missed" instead of triggering the durable-storage replay). See `docs/roadmap/10-bug-fixes.md` for full write-ups.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -224,6 +224,10 @@ Every DB query on game data needs a `where room_id = ?` clause. Every tRPC proce
 
 See [`docs/room-scoping.md`](docs/room-scoping.md) for the full rule with code examples, a code-review checklist, and a table of common failure patterns. Past bugs have been caused by missing room filters — treat this as a hard constraint, not a guideline.
 
+## Critical Architecture Rule: Concurrency, Timing, and Prompt Flows
+
+Fight pacing, the serialized engine lanes, `activeFlows`, and the interactive prompt lifecycle interact in non-obvious ways and have caused the most persistent production bugs (fights flying by, commands appearing ignored, multi-step flows crashing). Before touching `helpers/delay-times.ts`, `ring/index.ts` pacing, `events/room-event-bus.ts` prompts, or the server command pipeline in `trpc/router.ts`, read [`docs/engine-concurrency-and-timing.md`](docs/engine-concurrency-and-timing.md). Key invariants: interactive command actions run fire-and-forget in a per-`roomId:userId` lane (never room-wide — that starves other users); awaited workshop mutations must stay prompt-free; the `PROMPT_CANCELLED` sentinel must be translated to `PromptCancelledError` before reaching game code.
+
 ## Architecture Notes for New Connectors
 
 1. Implement the channel callback: `({ announce, question?, choices?, delay? }) => Promise`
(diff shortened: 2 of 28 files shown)


tweettypography:
All four review points addressed in 6c1e8a2:

1. **Replay race** — the live subscriber now attaches *before* the replay is assembled; events arriving during the async DB replay buffer in the queue, and duplicates of replayed events are dropped by id on drain. The `try/finally` now opens before the replay so a throw there can't leak the subscriber. `getEventsSinceForRingFeed` paginates internally (each page anchors on the previous page's last event id, so subsequent pages take the fast anchor path) up to a hard cap, and returns `limitReached` so the router emits a `system.gap` marker instead of presenting a truncated replay as complete.

2. **Retry-after-reset** — the writer's detach function now sets a cancellation flag that the retry loop checks between every step, so a write sitting in its backoff when `resetRoomState()` runs can no longer land afterwards. (A transaction already in flight at the instant of detach can still commit — the flag closes the seconds-long backoff window, which was the exploitable part.)

3. **Flow-lock race** — `activeFlows` is now a `Map<flowKey, commandId>`; each flow's `.finally()` only clears the lock if it still holds its own token, so a cancelled flow settling late can't release a newer command's lock. `cancelFlow`'s force-clear remains unconditional since that's explicit user intent. Covered by an end-to-end test of the exact A-cancelled → B-starts → A-settles sequence.

4. **Scope/docs** — PR title and body rewritten to cover the full change set (#15–#18, #20–#25 + review fixes); `CLAUDE.md` and `AGENTS.md` no longer list #15/#16 as active and now describe the actual root causes, with the still-open #26 (process-wide shop singleton) listed instead.

Coverage added for everything called out as untested: per-user command lanes, workshop fail-fast, cancellation overlap, quick-action emission, subscribe-before-replay ordering + dedup, replay pagination + cap gap marker, retry cancellation, named/cancelled item selection, and a full fight through the real non-skipped pacing path (those branches previously had zero coverage because the suite always sets `DECK_MONSTERS_SKIP_DELAYS`). 680 tests green across all packages.

---

